PR #40, opened 2017-01-12
Added argparse factory, working skip table

emattiza:
To make better, perhaps there is a way to not re-reflect?

Implementation performs two tasks:
* factory for arg parse parser
* logic to handle skipped tables
  * Currently runs on tables spec'ed or grabs all,
  * diffs skiptable args if specified from tables found
  * re-reflects schema from tables list less the skipped_tables
-----------------------------------------
Apologies for the wrong pull request earlier. This will be on the 1.1 branch exclusively.

Ref: Issue #39

### DIFF
--- a/sqlacodegen/main.py
+++ b/sqlacodegen/main.py
@@ -1,33 +1,59 @@
-""" """
-from __future__ import unicode_literals, division, print_function, absolute_import
+from __future__ import unicode_literals, division, print_function, \
+    absolute_import
 import argparse
 import codecs
 import sys
 
 from sqlalchemy.engine import create_engine
 from sqlalchemy.schema import MetaData
+import pkg_resources
 
 from sqlacodegen.codegen import CodeGenerator
-import sqlacodegen
+
+
+def parser_factory():
+    '''
+    A factory method to generate a parser with standard args
+    :return: ArgumentParser
+    '''
+    parser = argparse.ArgumentParser(
+        description='Generates SQLAlchemy model '
+                    'code from an existing database.')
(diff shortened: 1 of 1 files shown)


emattiza:
def _filter_tables_list(table_name, metadata_obj):
        '''
        A callable function passed to MetaData.reflect which returns boolean predicates on args entered
        First, return False for all tables in skiptables
        Second, if tables arg is non-empty, only return tables specified; Else return all tables
        :param table_name: string
        :param metadata_obj: sqlalchemy.MetaData
        :return: bool
        '''
        if table_name in args.skiptables:
            return False
        if args.tables:
            if table_name in args.tables:
                return True
            else:
                return False
        else:
            return True

This uses the sqlalchemy [metadata.reflect(only=)](http://docs.sqlalchemy.org/en/latest/core/metadata.html#sqlalchemy.schema.MetaData.reflect.params.only) in a callable which should prevent re-reflection and is a bit cleaner. I have this in a separate branch but will merge into my 1.1 and this pull request if you feel this is a better choice.

agronholm:
I don't quite understand the point here.

agronholm:
If your intent was on adding a "skip-tables" feature, I'm not inclined to merge something like that because it runs contrary to the intended workflow of sqlacodegen. If you want to skip some tables, why not just delete them from the generated code? Moreover, this PR touches a lot more than just one extra option and that is a no-no.

gmonkman:
Because sqlcodeagen does not support all data types, for example, there is no support for the MSSQL spatial data types. Being able to exclude tables which contain unsupported data type would be useful

PR #42, opened 2017-01-18
change numeric type to integer for auto_increment column in MSSQL

scymen:


agronholm:
Your PR is far too overreaching. Try narrowing down its scope.

Issue #48, opened 2017-06-15
Tables with no primary keys (confusingly) don't generate classes


ernstki:
This may seem patently obvious to SQLAlchemy veterans, but when `sqlacodegen` found tables missing primary keys and produced `Table` instances for these rather than proper classes, I was pretty flummoxed at first.

Fortunately I found a [couple](https://groups.google.com/d/msg/sqlalchemy/EARLfSP0Z6Y/H-XbvofTCQAJ) [posts](https://groups.google.com/d/msg/sqlalchemy/uQ7MijlHW1Y/3xaA7vJ6BwAJ) in the sqlalchemy Google Group that prompted me to have a closer look.

I get that tables with no primary keys of their own could just be relationship tables, or some other perfectly valid, unexceptional reason that's not worth making noise about when the script runs. But perhaps this should be mentioned in the README all the same?

PR #49, opened 2017-06-15
Add note to README about tables w/ no primary key

ernstki:
Add a note to the README stating that tables with no primary key will be generated as instances of `Table` rather than classes.

Ref: #48


agronholm:
A better solution would be to emit a warning in such cases (to stderr) after generating the code, wouldn't it?

ernstki:
I don't know how to rebase against upstream changes on the same pull request, so I'm closing this one. I'll look at adding some warnings to stderr once I've re-sync'd with your `master` branch.

PR #41, opened 2017-01-13
Updated links from Bitbucket to GitHub

neilsh:
Updated links in the README to reflect the migration from Bitbucket to GitHub

PR #50, opened 2017-09-06
update url in setup.py

waynenilsen:
I noticed this on pypi, thought you would want to change it?

I saw #41 this also changes the `setup.py` file.


agronholm:
Thanks.

Issue #60, opened 2018-05-10
Hyphen in table name generates invalid python


danohuiginn:
If I run sqlacodegen against a (mysql) table with a name like "foo-bar", it will output something like:

> foo-bar = Table(...)

which will generate a SyntaxError if you import it in python, because the hyphen is interpreted as an arithmetic operator.

From looking at the code, it seems that sqlacodegen.codegen.ModelClass#_tablename_to_classname does not call sqlacodegen.codegen._convert_to_valid_identifier, which would handle this case.

PR #61, opened 2018-05-10
Handle invalid characters in table names

danohuiginn:
Fixes Issue #60

Also remove some unused duplicated code


agronholm:
I see it now. Thanks!

PR #74, opened 2018-08-20
Option to import declarative base class from another module.

nijm:
Allows models split across separate files to share the same Base class and metadata.


agronholm:
I don't believe there is any value in this change. The user can just add the import after generating the model code. The point of sqlacodegen is not to produce picture perfect code for the user's specific needs, but to save them a lot of time which would otherwise have been consumed by manually typing the code for the models.

Issue #65, opened 2018-05-30
sqlalchemy.exc.CompileError: VARCHAR requires a length on dialect mysql


zhanghe06:
sqlacodegen==2.0

```
Traceback (most recent call last):
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/bin/sqlacodegen", line 11, in <module>
    sys.exit(main())
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlacodegen/main.py", line 51, in main
    args.noinflect, args.noclasses)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlacodegen/codegen.py", line 395, in __init__
    not nojoined)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlacodegen/codegen.py", line 149, in __init__
    super(ModelClass, self).__init__(table)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlacodegen/codegen.py", line 77, in __init__
    column.type = self._get_adapted_type(column.type, column.table.bind)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlacodegen/codegen.py", line 94, in _get_adapted_type
    if new_coltype.compile(bind.dialect) != compiled_type:
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlalchemy/sql/type_api.py", line 562, in compile
    return dialect.type_compiler.process(self)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlalchemy/sql/compiler.py", line 293, in process
    return type_._compiler_dispatch(self, **kw)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlalchemy/sql/visitors.py", line 81, in _compiler_dispatch
    return meth(self, **kw)
  File "/Users/zhanghe/code/news_flask_restful/news_flask_restful.env/lib/python2.7/site-packages/sqlalchemy/sql/compiler.py", line 2933, in visit_string
    return self.visit_VARCHAR(type_, **kw)
[...]

PR #75, opened 2018-08-20
Import vendor-specific types when generic types cannot be used.

### LINKED ISSUE #67: Generate models got wrong types
sqlalchemy: 1.2.8
sqlacodegen: 2.0.0

Backend: MySQL 5.6

Table:
```
+----------------------+-------------+------+-----+---------+----------------+
| Field                | Type        | Null | Key | Default | Extra          |
+----------------------+-------------+------+-----+---------+----------------+
| id                   | int(11)     | NO   | PRI | NULL    | auto_increment |
```

When generate models, the types used in the model are `INTEGER` instead of `Integer`.
```
from sqlalchemy import BIGINT, CHAR, Column, DateTime, Float, INTEGER,
```

results in the columns are:
```
id = Column(INTEGER(11), primary_key=True)
```

This will raise error as the `INTEGER` or `BIGINT` takes no params:
```
TypeError: object() takes no parameters
```
I have to reverted `sqlacodgen` to v1.1.6 to make it work again.

### LINKED ISSUE #70: When use code generate form sqlacodegen，a Error occur
sqlacodegen generate 
class Author(Base):
    __tablename__ = 'authors'

    author_id = Column(Integer(11), primary_key=True)
    author_name = Column(String(255))

when query：
objs = session.query(Author).all()

error：
    author_id = Column(Integer(11), primary_key=True)
TypeError: object() takes no parameters

nijm:
If a vendor-specific type cannot be adapted to a generic sqlalchemy type then import the vendor-specific type rather than the generic type. Some vendor-specific types have different constructors. E.g., MySQL INTEGER, which can take a display width.

Fixes #67 
Potentially fixes #70 (not enough info in issue to know for sure)

The code was careful not to adapt column types further if they rendered differently from the adapted type, then imported the generic types anyway when the imports were rendered.

Note: I changed the < 1.1 test_arrays ARRAY import from:
`from sqlalchemy.dialects.postgresql.base import ARRAY`

to:
`from sqlalchemy.dialects.postgresql import ARRAY`

### DIFF
--- a/sqlacodegen/codegen.py
+++ b/sqlacodegen/codegen.py
@@ -5,6 +5,7 @@ import inspect
 import re
 import sys
 from collections import defaultdict
+from importlib import import_module
 from inspect import ArgSpec
 from keyword import iskeyword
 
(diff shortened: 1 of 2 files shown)


nijm:
The second commit fixes #65 .

vietvudanh:
@nijm thank you for the fix.
@agronholm can you take a look at this?

agronholm:
I will take a look at this today.

PR #78, opened 2018-09-05
Add support for multi-file export

Havoc24k:
Added --multifile CLI option to support the export of each model/table to a separate file.


agronholm:
Sorry, but I don't really see the value here. Why would anyone want to separate their model automatically to multiple modules? At the very least it would not be a common case (seeing as nobody has ever asked for this feature).

PR #79, opened 2018-09-09
Added --notables hack

Morabaraba:
Added --notables to sqlacodegen.

Not the cleanest method, but it works for me.

Just wanted to share. If cleanup is need to get request into master let me know what.

### DIFF
--- a/sqlacodegen/codegen.py
+++ b/sqlacodegen/codegen.py
@@ -331,7 +331,7 @@ class CodeGenerator(object):
     def __init__(self, metadata, noindexes=False, noconstraints=False, nojoined=False,
                  noinflect=False, noclasses=False, indentation='    ', model_separator='\n\n',
                  ignored_tables=('alembic_version', 'migrate_version'), table_model=ModelTable,
-                 class_model=ModelClass,  template=None):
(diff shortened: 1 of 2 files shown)


agronholm:
Sorry, what is the point of this?

Morabaraba:
Wanted to generate code without the table declarations; just the orm classes in the generated py code.

Added a `--notables` to accomplish it. Feels a bit "hacky". But it does the job.

agronholm:
What about tables that do not have primary keys? And what about association tables? Are you just going to leave them out?

Morabaraba:
Those can be issue(s) one run into by using this `--notables` hack.

But I have not found a issue yet for my (simple) use case. I dynamically load the ORM classes dynamically to create [`ModelView`](https://flask-admin.readthedocs.io/en/latest/api/mod_contrib_sqla/) classes to browse my db tables in a crud way.

I found the table declaration creating conflict with my sqlalchmy metdata engine and wanted a way to remove it.

agronholm:
> I found the table declaration creating conflict with my sqlalchmy metdata engine and wanted a way to remove it.

That simple way would be to edit the generated models file and remove the tables you don't want.

Sorry, I cannot accept this PR.

PR #80, opened 2018-09-20
Allow giving connect args for db engine

joystein:
For instance:

sqlacodegen --connectarg read_default_file=~/.my.cnf --connectarg read_default_group=mysql_app mysql+mysqldb://127.0.0.1/db

would allow keeping login credentials in a common place for mysql.

### DIFF
--- a/sqlacodegen/main.py
+++ b/sqlacodegen/main.py
@@ -28,6 +28,7 @@ def main():
     parser.add_argument('--noclasses', action='store_true',
                         help="don't generate classes, only tables")
     parser.add_argument('--outfile', help='file to write output to (default: stdout)')
+    parser.add_argument('--connectarg', action='append', help="add connect args for engine")
     args = parser.parse_args()
 
     if args.version:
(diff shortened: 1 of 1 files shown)


joystein:
@agronholm : have you considered the changes in this pull request? I submitted this very same patch before on bitbucket, but then you wanted to wait until version 2.

agronholm:
Ah, sorry. Two changes I will need:

1. No need for the conditional – you could have a default value of an empty list for connectargs and do the parameter split in any case.
2. Add proper documentation on how to use this new option.

joystein:
When looking back at this old code I realize that I have probably made a mistake in believing that I couldn't pass these connect arguments as part of the url on the command line, probably due to '&' not being properly escaped and thereby causing the command to fail when the first part of it is sent to background and the rest is ignored. With proper escaping things works as expected with url params.

Being able to explicitly pass these connect arguments as command line options to sqlalcodegen could be a nice convenience, but is no longer a blocker for me.

Sorry for the noise. For my part this can be closed.

PR #82, opened 2018-10-12
Added support for env var name as db url

hvgab:
I use this tool with pipenv.
If my db url is in my system/user envs I can do:
`pipenv run sqlacodegen --tables Employees %DB_URL%`
But if my DB_URL is inside a .env file that is loaded by pipenv this does not get translated to the actual value.

This pull just adds a try to get env var named `args.url`, if it does not exist, assume args.url is a normal url.

### DIFF
--- a/sqlacodegen/main.py
+++ b/sqlacodegen/main.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals, division, print_function, absolute_impo
 import argparse
 import io
 import sys
+import os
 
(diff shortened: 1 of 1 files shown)


agronholm:
This does not look like something I would like to add to the codebase. Do all other tools support something like this?

agronholm:
Doesn't seem very important since this has gone for 3½ years without a reply.

Issue #57, opened 2018-01-17
Impossible to generate schema with classes only (opposite of --noclasses)


thirdeye-hichame:
As discussed previously on [stackoverflow](https://stackoverflow.com/a/35916967), some users would like to generate schemas with all tables represented as SQLAlchemy classes.

Would it be possible to add an `--onlyclasses` argument to sqlacodegen?

Many thanks

agronholm:
For what purpose would you want association tables to be classes instead of interim tables in many-to-many relationships? As for the other case, tables without a pk column cannot be classes anyway.

thirdeye-hichame:
@agronholm , thanks for the response. I wanted to generate classes similar to [here](http://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html#association-object)

Since it was 5 months ago I don't have the exact specific use case any more but I know I wanted to insert objects into an interim table with a composite primary key `PRIMARY KEY(parent_id, children_id)`

To solve this problem I forked `agronholm/sqlacodegen`  and added the argument myself.

agronholm:
I guess I could add something like `--nomanytomany` which would skip the detection of association tables, in which case they would be rendered as classes instead of Tables (assuming they have primary keys). Would this be acceptable?

sam-alloy:
I second the idea to have an option like `forceclasses`.

According to [the doc](https://docs.sqlalchemy.org/en/13/faq/ormconfiguration.html)

> It’s important to note that we’re only talking about the SQLAlchemy ORM; an application which builds on Core and deals only with Table objects, select() constructs and the like, does not need any primary key to be present on or associated with a table in any way (though again, in SQL, all tables should really have some kind of primary key, lest you need to actually update or delete specific rows).

In our application it is exactly the use case, where we don't use ORM but just want to have proper type check.

petergok:
This would be very useful for us to be able to generate classes we can query for tables that don't have a primary key. Here's a quick proposed solution with an `--onlyclasses` argument: https://github.com/perpetua1/sqlacodegen

agronholm:
SQLAlchemy ORM classes do not work without a primary key definition: https://docs.sqlalchemy.org/en/14/faq/ormconfiguration.html#how-do-i-map-a-table-that-has-no-primary-key
You do know that you can use a Session to query plain tables too?

petergok:
[...]

PR #83, opened 2018-11-06
command line arg 'notables', logic, readme update

aktivkohle:
I have created some code and added a line to the readme to address this issue:
https://github.com/agronholm/sqlacodegen/issues/57

It has been done in such a way that if no `--notables` argument is given, everything should run as it usually does. In ran the tests and got the same result on this fork and the master fork.

If any additional print statements or readme additions are needed to warn the user about using the `--notables` argument, I can add them, or write any additional tests.


agronholm:
It's been a long while but I don't think sqlacodegen should produce nonworking classes (without primary keys). That would just generated a bunch of new bug reports.

PR #92, opened 2019-04-21
support table/column comment

yihuang:


sqlacodegen/codegen.py
```diff
@@ -558,6 +558,8 @@ def render_column(self, column, show_name):
         elif has_index:
             column.index = True
             kwarg.append('index')
+        if column.comment:
```
agronholm: Separate with one empty line to distinguish the block from the previous if...else.

sqlacodegen/codegen.py
```diff
@@ -612,6 +614,9 @@ def render_table(self, model):
         if model.schema:
             rendered += "{0}schema='{1}',\n".format(self.indentation, model.schema)
 
+        if model.table.comment:
+            rendered += f'{self.indentation}comment="{model.table.comment}",\n'
```
agronholm: Do not use f-strings as they require Python 3.7+.

PR #97, opened 2019-06-16
add comment support

koalas8:
Comment will be generated by default. You can use --nocommits to make it ineffective


agronholm:
The PR required so many changes that I decided it would be easier to make my own changeset. I've credited you for your work in the changelog however.

Issue #87, opened 2019-01-28
KeyError: 'MYCOLUMN '


lucasrcezimbra:
I receive a KeyError running 

```
$ sqlacodegen --schema=myschema --tables=mytable 'oracle://myuser:mypassword@myhost'

Traceback (most recent call last):
  File "/home/lucas/workspace/notebooks/.venv/bin/sqlacodegen", line 11, in <module>
    sys.exit(main())
  File "/home/lucas/workspace/notebooks/.venv/lib/python3.6/site-packages/sqlacodegen/main.py", line 51, in main
    args.noinflect, args.noclasses)
  File "/home/lucas/workspace/notebooks/.venv/lib/python3.6/site-packages/sqlacodegen/codegen.py", line 403, in __init__
    if isinstance(table.c[colname].type, String):
  File "/home/lucas/workspace/notebooks/.venv/lib/python3.6/site-packages/sqlalchemy/util/_collections.py", line 198, in __getitem__
    return self._data[key]
KeyError: 'MYCOLUMN '
```
```
$ sqlacodegen --schema=myschema --tables=mytable 'oracle://myuser:mypassword@myhost'

Traceback (most recent call last):
  File "/home/lucas/workspace/notebooks/.venv/bin/sqlacodegen", line 11, in <module>
    sys.exit(main())
  File "/home/lucas/workspace/notebooks/.venv/lib/python3.6/site-packages/sqlacodegen/main.py", line 51, in main
    args.noinflect, args.noclasses)
  File "/home/lucas/workspace/notebooks/.venv/lib/python3.6/site-packages/sqlacodegen/codegen.py", line 403, in __init__
    if isinstance(table.c[colname].type, String):
  File "/home/lucas/workspace/notebooks/.venv/lib/python3.6/site-packages/sqlalchemy/util/_collections.py", line 198, in __getitem__
    return self._data[key]
KeyError: 'mycolumn '
```

I fixed that changing [this line](https://github.com/agronholm/sqlacodegen/blob/master/sqlacodegen/codegen.py#L401) to:

```
colname = _re_column_name.match(match.group(1)).group(3).lower().strip()
```

Does this correction make sense? Should I open a pull request with this?

pedrozan:
Same thing happened to me, your fix worked!

Lrcezimbra:
You can use [this](https://github.com/Lrcezimbra/sqlacodegen) with
```
[...]

PR #100, opened 2019-07-10
fix: Resolve issues with oracle column names not matching because of case and whitespace

cjpit:


agronholm:
Please explain in more detail what the problem is you're trying to fix here. I will require at least a new test added to verify that the patch fixes the problem.

softwarepk:
This and pull request #105 are both trying to solve the same issue #87.

agronholm:
It's been 9 months without those tests I asked for so I'm going to reject this.

PR #101, opened 2019-07-16
Added '--nojson' option and generate asdict(), __repr__, __str__ .

nesiax:
Hi, this patch add a '--nojson' option to NOT generate __repr__ and __str__ and asdict() methods for generated classes.
I haven't tested it too much except for the PostgreSQL database located at:
http://www.postgresqltutorial.com/postgresql-sample-database/
So, you probably want to change it to --json if you want to NOT generate those methods and only do that on purpose, the idea behind that is you can do testing using the python command line interpreter with the results comming from SqlAlchemy as explained in the Tip from https://docs.sqlalchemy.org/en/13/orm/tutorial.html#declare-a-mapping.
"The User class defines a __repr__() method, but note that is optional; we only implement it in this tutorial so that our examples show nicely formatted User objects."

### DIFF
--- a/sqlacodegen/codegen.py
+++ b/sqlacodegen/codegen.py
@@ -320,16 +320,23 @@ class ManyToManyRelationship(Relationship):
 
 class CodeGenerator(object):
     template = """\
-# coding: utf-8
 {imports}
 
 {metadata_declarations}
 
 
-{models}"""
+{models}
+
+
+# For Emacs:
+# Local Variables:
+# mode: python
+# coding: utf-8
+# End:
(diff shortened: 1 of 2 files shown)


softwarepk:
This and pull request #105 are both trying to solve the same issue #87.

agronholm:
I'm not inclined to accept this. This is someone's specific solution for serializing sqlalchemy model instances. It does not belong in a generic code generator. Moreover, this patch contains many changes unrelated to the actual feature, plus it doesn't add any tests.

PR #76, opened 2018-08-20
Allow custom naming of class models and relationships.

nijm:
Pass class naming and relationship naming functions as cli args. See the tests for examples.

The class naming function takes the table name and returns a class name.

The relationship naming function takes a ForeignKeyConstraint object and a relationship type (one-to-one/one-to-many/many-to-many) and returns a tuple of (relationship_name, backref_name).

I'm sure it's fairly common for naming conventions to differ from the snake_case that sqlacodegen expects, especially since it's used for generating models from existing dbs, which the dev may not be able to change.

I'll happily take suggestions for improvements if you're concerned about how this is implemented.

PR #112, opened 2020-04-17
Add support for tablename prefix and custom naming functions

softwarepk:
In many legacy databases it is quite customary to have table names prefixed with some prefix. There are many other reasons for which such prefixes are used. For example it is quite normal to see WordPress database tables prefixed with wp_
When code is generated for such database tables, the prefix becomes part of the class name. Which in fact is not desirable. This update allows us to ignore the specified prefix in generating the class names.

Example: **wp_** is a prefix to ignore.
If table name is **wp_user**, the standard code generation will generate class name **WpUser**. The modified code when passed with command line parameter ** --tableprefix  wp_** will generate class name **User**

### **Update:**
This pull request now incorporates the custom class naming functions feature implemented by pull request #76.

### DIFF
--- a/sqlacodegen/codegen.py
+++ b/sqlacodegen/codegen.py
@@ -8,7 +8,6 @@ from collections import defaultdict
 from importlib import import_module
 from inspect import ArgSpec
 from keyword import iskeyword
-from enum import Enum as PyEnum
 
 import sqlalchemy
 import sqlalchemy.exc
@@ -257,7 +256,7 @@ class ModelClass(Model):
             child.add_imports(collector)
 
 
-class RelationshipType(PyEnum):
+class RelationshipType():
     ONE_TO_ONE = "ONE_TO_ONE"
     MANY_TO_ONE = "MANY_TO_ONE"
     MANY_TO_MANY = "MANY_TO_MANY"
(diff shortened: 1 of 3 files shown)


softwarepk:
This pull request now incorporates the custom class naming functions feature implemented by pull request #76.

Check failing because Python3.x isn't supported here.

softwarepk:
@agronholm for review

mzpqnxow:
I see this is related to #76 where I also left a note to the effect of "does anyone have time to clean up the merge conflicts to finish this?"

Thanks!

mzpqnxow:
@agronholm do you think this is something that you will accept? I've been using a branch with it on a large project using constraints.txt to point to my branch, but it would be great to have in upstream

agronholm:
I am currently in the middle of a major refactoring effort, part of which is intended to provide more customizability. When that is done, I would have you check if it's dynamic enough for your purposes. If not, I can always make it even more customizable.

agronholm:
You should check out 3.0.0b1 and see if the refactored code works for your purposes. Let me know!

agronholm:
At any rate, I'm closing this PR as redundant.

Issue #71, opened 2018-07-23
generates models with invalid identifiers


bwduncan:
When generating a model from a MSSQL database, sqlacodegen generates lines like:

`    t_PG_summaryView-Jai = Table(`

which is a syntax error because the identifier doesn't match [A-Za-z0-9_].

PR #113, opened 2020-04-21
Fix for issue #71

softwarepk:
Issues with table name identifiers getting generated with invalid characters has been fixed. 
Identifiers will now match the [A-Za-z0-9_] specs.

### DIFF
--- a/sqlacodegen/codegen.py
+++ b/sqlacodegen/codegen.py
@@ -158,8 +158,22 @@ class Model(object):
             if len(index.columns) > 1:
                 collector.add_import(index)
 
+    @staticmethod
+    def _convert_to_valid_identifier(name):
+        assert name, 'Identifier cannot be empty'
+        if name[0].isdigit() or iskeyword(name):
+            name = '_' + name
+        elif name == 'metadata':
+            name = 'metadata_'
+
+        return _re_invalid_identifier.sub('_', name)
+
 
 class ModelTable(Model):
+    def __init__(self, table):
+        super(ModelTable, self).__init__(table)
+        self.name = self._convert_to_valid_identifier(table.name)
+
     def add_imports(self, collector):
         super(ModelTable, self).add_imports(collector)
         collector.add_import(Table)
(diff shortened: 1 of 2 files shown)


softwarepk:
@agronholm for review

sqlacodegen/codegen.py
```diff
@@ -158,8 +158,22 @@ def add_imports(self, collector):
             if len(index.columns) > 1:
                 collector.add_import(index)
 
+    @staticmethod
+    def _convert_to_valid_identifier(name):
```
agronholm, line 162: Why did you move the method here? It doesn't seem to be used outside of ModelTable even in your version.

softwarepk: Previously **_convert_to_valid_identifier** was only used in ModelClass and it was a static method of this class. Now **_convert_to_valid_identifier** is also used in ModelTable so it is moved to the base class.

sqlacodegen/codegen.py
```diff
@@ -158,8 +158,22 @@ def add_imports(self, collector):
...
+        elif name == 'metadata':
+            name = 'metadata_'
+
+        return _re_invalid_identifier.sub('_', name)
+
 
 class ModelTable(Model):
+    def __init__(self, table):
+        super(ModelTable, self).__init__(table)
+        self.name = self._convert_to_valid_identifier(table.name)
```
softwarepk, line 175: **_convert_to_valid_identifier** used in ModelTable

sqlacodegen/codegen.py
```diff
@@ -208,16 +222,6 @@ def _tablename_to_classname(cls, tablename, inflect_engine):
...
-        assert name, 'Identifier cannot be empty'
-        if name[0].isdigit() or iskeyword(name):
-            name = '_' + name
-        elif name == 'metadata':
-            name = 'metadata_'
-
-        return _re_invalid_identifier.sub('_', name)
-
     def _add_attribute(self, attrname, value):
         attrname = tempname = self._convert_to_valid_identifier(attrname)
```
softwarepk, line 226: **_convert_to_valid_identifier** used in ModelClass

sqlacodegen/codegen.py
agronholm: Ah, my bad – the diffs can sometimes be confusing and my memory of the code layout is less than perfect :)

agronholm:
Thanks!

softwarepk:
Thanks!

Issue #108, opened 2020-01-03
Implement bi-directional relationships


t-o-o-m:
Creating a bidirectional many-to-many relationship seems to be impossible with sqlacodegen.

# Desired model
```
association_table = Table('association', Base.metadata,
    Column('left_id', Integer, ForeignKey('left.id')),
    Column('right_id', Integer, ForeignKey('right.id'))
)

class Parent(Base):
    __tablename__ = 'left'
    id = Column(Integer, primary_key=True)
    children = relationship(
        "Child",
        secondary=association_table,
        back_populates="parents")

class Child(Base):
    __tablename__ = 'right'
    id = Column(Integer, primary_key=True)
    parents = relationship(
        "Parent",
        secondary=association_table,
        back_populates="children")
```

# Model created by sqlacodegen
```
association_table = Table('association', Base.metadata,
    Column('left_id', Integer, ForeignKey('left.id')),
    Column('right_id', Integer, ForeignKey('right.id'))
)

class Parent(Base):
    __tablename__ = 'left'
    id = Column(Integer, primary_key=True)
    children = relationship("Child",
                    secondary=association_table)

class Child(Base):
    __tablename__ = 'right'
    id = Column(Integer, primary_key=True)
```

A single backref definition in the parent would suffice according to https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html?highlight=relationship#basic-relationship-patterns

osmnorhn:
Is there any updates regarding to this situation ?
Is there any plans to support bi-directional relationships ?

adyah-cisco:
Will this be fixed with the PR that I see? Would love to have this implemented.

agronholm:
I will try to add it in v3.0. Most of the work for that version is done, so the process of adding features can now resume.

agronholm:
I have code for this now, still 13 failing tests but getting there. It will be in v3.0.

agronholm:
Oh, and I'm making it the default.

agronholm:
3 failing tests left. Problems are focused on one-to-many relationships which are either self referential or have multiple foreign keys between the source and target tables.

sdementen:
The issue is closed yet as my request is tightly linked I post it here (I can make a new issue if you prefer)
Would it be possible to have both sidea of the relationship to be created so that IDEs can provide autocompletion for both sides?

PR #121, opened 2020-07-13
Support creating bidirectional relationships

Chrisyunhua:
Support creating bidirectional relationships and adding `back_populates` to `relationship`.

related issue: #108


agronholm:
Thanks for the effort, but the changes are unfortunately incompatible with the codebase after the major refactoring that just went in, so I'm going to have to turn this down. The feature itself is something I will attempt to add to v3.0.

Issue #124, opened 2020-09-25
[HAS PR] columns of type citext are recognized as NullType (PostgreSQL)


mzpqnxow:
When using the `citext` (case-insensitive text) extension with PostgreSQL, `sqlacodegen` doesn't recognize the column type and emits a `NullType` for the column in the output.

The (non-fatal) warning message at runtime is:

```
/venv/lib/python3.7/site-packages/sqlalchemy/dialects/postgresql/base.py:3015: SAWarning: Did not recognize type 'citext' of column 'registrar_technical_contact_email'
  "Did not recognize type '%s' of column '%s'" % (attype, name)
```

This is easy to fix manually after the model is emitted by `sqlacodegen` by simply replacing the `NullType` instances with `Text`. At this point, it works as expected

If I understand correctly, the `citext` extension is very primitive and a bit of a hack- it just causes all operators to use `lower()` on the values when they are of type `citext`. It's probably a better practice to use check constraints with PostgreSQL to accomplish something roughly equivalent to what `citext` does, but it's a convenient way to do this without adding check constraints

Would it be appropriate to add explicit `citext` support to `sqlacodegen`? Even if all it does is emit `Text`? Hand editing it isn't a horrible task, but it would be nice if it emitted `Text` in the first place

Thanks, and thank you for the effort in making such an excellent tool!

mzpqnxow:
Adding one additional note to this- it's simple to fix the Null columns to `CIText` after running `sqlacodegen`, but it's also very easy to make a mistake and not realize that a `Null` column is a `CIText[]` type as opposed to `CIText[]`. This is the expected behavior because `CIText` is not explicitly supported (and therefore a `CIText[]` is meangingless to `sqlacodegen`) but I'm pointing it out because I've made this mistake a few times now

mzpqnxow:
I added a PR for this, #125 which implements this as an extra

PR #125, opened 2020-10-10
Add extra 'citext' to add support for CIText  and CIText arrays on PostgreSQL database

mzpqnxow:
After entering #124, I took a quick look at this and saw that it was pretty trivial to add. I added it as an optional / extra so it can be installed using `pip install sqlacodegen[citext]` without impacting users that don't want or need it

Is this something you'll take upstream?

Thanks!

### DIFF
--- a/sqlacodegen/codegen.py
+++ b/sqlacodegen/codegen.py
@@ -19,12 +19,6 @@ from sqlalchemy.sql.sqltypes import NullType
 from sqlalchemy.types import Boolean, String
 from sqlalchemy.util import OrderedDict
 
-# Support CIText and CIText[] in PostgreSQL via sqlalchemy-citext
-try:
-    from citext import CIText  # noqa: F401
-except (NameError, ImportError):
-    pass
-
 # The generic ARRAY type was introduced in SQLAlchemy 1.1
 try:
     from sqlalchemy import ARRAY
@@ -40,6 +34,12 @@ except ImportError:
 # Conditionally import Geoalchemy2 to enable reflection support
 try:
     import geoalchemy2  # noqa: F401
+except ImportError:
+    pass
+
+# Support CIText and CIText[] in PostgreSQL via sqlalchemy-citext
+try:
+    import citext  # noqa: F401
 except ImportError:
     pass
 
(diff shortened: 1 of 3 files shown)


agronholm:
This should be mentioned in the docs or otherwise people won't know about the extra.

mzpqnxow:
@agronholm good point, thank you. I went ahead and added it

sqlacodegen/codegen.py
```diff
@@ -19,6 +19,12 @@
 from sqlalchemy.types import Boolean, String
 from sqlalchemy.util import OrderedDict
 
+# Support CIText and CIText[] in PostgreSQL via sqlalchemy-citext
+try:
+    from citext import CIText  # noqa: F401
+except (NameError, ImportError):
```
agronholm, line 25: This should be moved to after the Geoalchemy2 import.

Under what circumstances can this raise NameError?

sqlacodegen/codegen.py
```diff
@@ -19,6 +19,12 @@
 from sqlalchemy.types import Boolean, String
 from sqlalchemy.util import OrderedDict
 
+# Support CIText and CIText[] in PostgreSQL via sqlalchemy-citext
+try:
+    from citext import CIText  # noqa: F401
```
agronholm, line 24: Is it not enough to import the module?

sqlacodegen/codegen.py
mzpqnxow: Oops, that was supposed to be `ModuleNotFoundError` not `NameError`. Though I realize now that's overly specific and will fail on Python2 anyway as it doesn't have `ModuleNotFoundError`- which is a subclass of `ImportError`. Will remove that and commit, nice catch

sqlacodegen/codegen.py
mzpqnxow: Should be no different since it's not directly used. I'll change that to only import the module

sqlacodegen/codegen.py
mzpqnxow: (fixing the order as well)

mzpqnxow:
Pushed those changes up. Sorry 'bout that, especially the one that would have broken Python2 (!!!)

agronholm:
Thanks!

PR #133, opened 2021-04-28
Require setuptools sqlacodegen/main.py imports pkg_resources

hrnciar:


agronholm:
This is obsolete now that `importlib.metadata` is used instead. There is a conditional dependency for the backport in place.

PR #142, opened 2021-06-26
added parameter for table name prefix

mk8310:


agronholm:
This use case is now covered by `TablesGenerator.generate_model_name()`.

PR #143, opened 2021-06-26
Enhancements for compatibility with SQLAlchemy >= 1.4 (warnings) and >= 2.0 (hard fail)

### LINKED ISSUE #132: declarative_base moved between SQL Alchemy 1.3.x and 1.4.x
sqlacodegen supports SQL Alchemy 1.3.x, but if you are on 1.4.x and want to try sqlacodegen, note declarative_base has been moved between 1.3.x and 1.4.x, so sqlacodegen's output model must be changed from 

```
from sqlalchemy.ext.declarative import declarative_base

```

to

```
from sqlalchemy.orm import declarative_base
```

Please see https://docs.sqlalchemy.org/en/14/orm/mapping_api.html#sqlalchemy.orm.declarative_base

mzpqnxow:
This is one way to fix #132. I admit it's not so pretty, but it does silence the warning and seems to work for me. It needs a review and some more testing (plus feedback on the approach and style)

### DIFF
--- a/src/sqlacodegen/codegen.py
+++ b/src/sqlacodegen/codegen.py
@@ -88,8 +88,9 @@ class ImportCollector(OrderedDict):
 
         self.add_literal_import(pkgname, type_.__name__)
 
-    def add_literal_import(self, pkgname: str, name: str) -> None:
-        names = self.setdefault(pkgname, set())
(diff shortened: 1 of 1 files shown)


mzpqnxow:
A fix for this is already implemented, closing

PR #152, opened 2021-08-24
Fix bug in foreign_keys generation

### LINKED ISSUE #151: Invalid foreign_keys generated for bidi relationship
This bug was introduced in https://github.com/agronholm/sqlacodegen/commit/ba08a71e2c85207ffedee52a6541e9b7df07f9d4.

The test code in `TestDeclarativeGenerator.test_onetomany_multiref` https://github.com/agronholm/sqlacodegen/blob/8956ac25de08cb57a68a9eb004647a5933eb730f/tests/test_generators.py#L927 actually seems to be invalid, at least with `sqlalchemy` 1.4.23 and python 3.9 (that said, the quotes from the docs above was written 7 years ago, so I suspect it's not a recent regression). If I copy the code snippet tested against to a separate file and run `print(sqlalchemy.select(SimpleItems))`, I get the following error:
`sqlalchemy.exc.ArgumentError: Column expression expected for argument 'foreign_keys'; got 'SimpleItems.parent_container_id'.`

Swapping `'[SimpleItems.parent_container_id]'` for `['SimpleItems.parent_container_id']` fixes the error.

From [the docs](https://docs.sqlalchemy.org/en/14/orm/join_conditions.html#handling-multiple-join-paths):
> When specifying foreign_keys with Declarative, we can also use string names to specify, however it is important that if using a list, the list is part of the string:
> `billing_address = relationship("Address", foreign_keys="[Customer.billing_address_id]")`

PR on the way!

laurentS:
Fixes #151

The test code used when checking bidirectional relationship generation is invalid, at least with sqlalchemy 1.4.x. Instead of a list of strings, a "string of list" is expected.
This PR corrects the test code, and the declarative generator to actually pass the test.

I added a separate function for rendering foreign keys as I'm not sure if other attributes need to be modified the same way (the docs don't mention anything about them).


Babboozeld:
I have the same issue, PR seems to fix problem looking at [sqlalchemy doc](https://docs.sqlalchemy.org/en/14/orm/join_conditions.html#handling-multiple-join-paths).

agronholm:
I have fixed the CI test automation since this PR was made. Would you mind merging master to this in order to trigger the test workflow?

laurentS:
@agronholm merged! just need your approval to get CI to run.

agronholm:
The code works, the tests are fine...I'm just having some concerns about the fact that it changes the API. I'm trying to think of some alternate way to fix this.

laurentS:
> The code works, the tests are fine...I'm just having some concerns about the fact that it changes the API. I'm trying to think of some alternate way to fix this.

If you want to keep the same API, or at least not break existing reuses of it, I guess you could take the behaviour of `render_foreign_keys` and move it into `render_column_attrs` to which you'd add an optional arg which defaults to the current behaviour?

Something like:
```python
def render_column_attrs(column_attrs: List[ColumnAttribute], foreign_key: Optional[bool] = False) -> str:
    if foreign_key:
        # do what render_foreign_keys does
    else:
        # current behaviour
```

Would that work better?

As a side question, this is a local function in a method of the generator, is it meant to be public API?

agronholm:
> As a side question, this is a local function in a method of the generator, is it meant to be public API?

The generators are designed to be customizable, so this will be part of the public API in v3.0.

agronholm:
Sorry, I must've been tired when I wrote those last comments. No, these are not intended to be part of the API, so what you have here is totally fine.

agronholm:
Thanks!

PR #153, opened 2021-08-26
Sort multicolumn indexes in table args

laurentS:
I am not exactly sure why this is happening, but I've had some cases where the output was inconsistent across machines (despite using the exact same postgresql version and code). Specifically, the order of multicolumn indexes in the Declarative generator was changing between machines.
As we use the generator repeatedly to sync our db schema definitions with a separate db migration system, it's important to us to have consistent results (we actually run CI checks on it).

This PR sorts the indexes by name so that the output is consistent across runs.

```python
class MyTable(Base):
    __tablename__ = 'mytable'
    __table_args__ = (
    # the following 2 lines change order between executions. This PR sorts them by name for consistency
        Index('mytable_created_at__state_idx_composite_btree', 'created_at', 'state'),
        Index('mytable_score_updated_at_idx', 'score', 'updated_at'),
        {'schema': 'public'}
    )
```


laurentS:
I've ended up adding the same fix to the tables generator as we eventually hit the same problem with it.

agronholm:
I suppose there is no harm in adding this.

Issue #154, opened 2021-08-26
Missing MetaData import with declarative generator


laurentS:
With this table `create table codegen.mytable (name text);` on postgresql 12 (and no "Model" table in the schema), if I run the `declarative` generator, I get the following code (with `sqlacodegen` 8956ac2):

```python
from sqlalchemy import Column, Table, Text
from sqlalchemy.orm import declarative_base

metadata = MetaData()   # the import is missing


t_mytable = Table(
    'mytable', metadata,
    Column('name', Text),
    schema='codegen'
)
```

benman1:
Hi. I think this is a great project. I created something for myself that's using Table as output, but it's far less flexible with the output. 

This works from python (maybe something like this could be in the Readme?):
```python
from sqlalchemy import MetaData, Table, Column, Integer, ForeignKey, MetaData
from sqlacodegen.generators import DeclarativeGenerator

metadata = MetaData()
Table('user_order', metadata,
     Column('id', Integer, primary_key=True),
     Column('user_id', 
))
engine = create_engine(
  "bigquery://whatever",
)  # not sure if there's a way to not use an engine if all I want is convert from table to base
generator = DeclarativeGenerator(metadata, bind=engine, options=[])
print(generator.generate())
```

Output from this is as expected:
```python
from sqlalchemy import Column, Integer
from sqlalchemy.orm import declarative_base
from sqlalchemy.sql.sqltypes import NullType

Base = declarative_base()


class UserOrder(Base):
    __tablename__ = 'user_order'

    id = Column(Integer, primary_key=True)
    user_id = Column(NullType)

```

But if I actually point it at bigquery or give it real tables, there are no error message, just `Table` style output.

votti:
@benman1 I think this should be a separate issue - but I have run into this myself, so I want to quickly document how to solve this:
The issue is that `bigquery` does not have primary key constraints and the default behaviour of `sqlacodegen` is to not use a declarative model for tables without primary key:
https://github.com/agronholm/sqlacodegen/blob/31583dace790f4b69b98b31576ac6ddd99155ee9/src/sqlacodegen/generators.py#L722-L726

So in order to get a proper model, your tables should have primary keys. As this concept is not really supported in bigquery, one quick and dirty workaround is to eg add a comment for the `primary key` and then modify this in the metadata before generating the model:

```
import io
import sys
from sqlalchemy import create_engine, MetaData, PrimaryKeyConstraint
[...]

PR #155, opened 2021-08-26
Add missing metadata import

laurentS:
This PR tries to fix https://github.com/agronholm/sqlacodegen/issues/154.
I am not sure this is the correct/best way of doing it, but it solves my problem, and the test suite seems happy with it.

Let me know if I should change the approach. And I should probably add a test as well :)


agronholm:
I've fixed this in another manner in 29614969.

Issue #122, opened 2020-08-27
Auto Increment Seq not created


matthewbal:
I used this amazing library to auto create a bunch of models from a huge Django Postgres database and it worked amazingly on the first try.
The only thing it missed was the auto sequences.

On every table, it correctly generated the "id" with the following definition (note, there are many non-django tables in this project too, these all have the same issue)

```
class DjangoMigration(Base):
    __tablename__ = 'django_migrations'

    id = Column(Integer, primary_key=True, server_default=text("nextval('django_migrations_id_seq'::regclass)"))
```

Unfortunately when I went to use this model to create a new table, I get the following issue:

```
E   sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "django_migrations_id_seq" does not exist
```

The workaround is to add the following into each table definition:
```
class DjangoMigration(Base):
    __tablename__ = 'django_migrations'

    django_migrations_id_seq = Sequence('django_migrations_id_seq', metadata=Base.metadata)

    id = Column(Integer, primary_key=True, server_default=text("nextval('django_migrations_id_seq'::regclass)"))
```

It'd be great if this library had auto created those lines as well! This could be user error, but I thought I'd report it in case there's a bug present.

Using:
Python 3.6.11 (official docker image)
postgres:11 (official docker image)
sqlacodegen 2.3.0 (pypi)

psycopg2==2.8.5
SQLAlchemy==1.3.19
SQLAlchemy-Utils==0.36.8

AntonOfTheWoods:
Given that it appears that such sequences get created automatically in many cases when a column is a singular primary key of type `Integer`, I just did a replace of `/, server_default.*)")//`. Might there be some drawbacks to this approach?

agronholm:
It's pretty difficult for sqlacodegen to determine if sequences should be declared explicitly or not. In this particular case I'm guessing that you used the generated metadata to create a new database and that's why you had the problem?

This could be added as an option I suppose.

agronholm:
Are sequences even reflected? How can I find them in the metadata?

PR #157, opened 2021-08-27
Included auto-increment to column generation

myinisjap:
This is to fix the issue in #122
I have tested this with both mysql and postgres.
The output from postgres is the following which matches what was originally expected.
```
class DjangoMigrations(Base):
    __tablename__ = 'django_migrations'

    id = Column(Integer, primary_key=True, autoincrement=True, server_default=text("nextval('django_migrations_id_seq'::regclass)"))
```

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -334,6 +334,8 @@ class TablesGenerator(CodeGenerator):
             kwargs['primary_key'] = True
         if not column.nullable and not is_sole_pk:
             kwargs['nullable'] = False
+        if column.autoincrement == True:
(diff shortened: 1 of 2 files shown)


agronholm:
Adding "autoincrement" doesn't fix your issue. Single column integer primary keys already have autoincrement enabled by default. So long as sequences are not reflected, they cannot be autogenerated in code by sqlacodegen. This problem has to be resolved upstream by enabling sequence reflection first.

TantumNTatari:
I hope it is okay to still ask about this. I am not sure I understand your explanation here. Could you explain more about what the underlying issue is with including this autoincrement option in the generator as proposed?

PR #158, opened 2021-08-30
2.3.0 with proxity

ahardjasa:
Attempting to update this to the upstream's 2.3.0 tag, which adds support for computed (`generated` in Postgres) columns.


ahardjasa:
Apologies, I did this backwards...

PR #188, opened 2022-04-01
Force pk

KhirilovA:
## What?
I've added an **option** _"forcepk"_ for **DeclarativeGenerator**.
If there is no primary key it forces it on the **first column** and makes a model
## Why?
This change may be helpful for people who want to get a model out of **view**
## How?
This includes an added option _"forcepk"_ in **valid_options**, and if else statement for chekcing whether it has primary key or not.
With **append_constraint** method it forces pk on first column, if such option was chosen
```
if 'forcepk' in self.options:
    if not table.primary_key:
        first_col_name = table.c.values()[0].name
        pk = PrimaryKeyConstraint(first_col_name)
        table.append_constraint(pk)
```
## Testing
I haven't added tests for it yet, but it won't be much a problem to do them

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -591,18 +591,32 @@ class DeclarativeGenerator(TablesGenerator):
                 links[tablename].append(model)
                 continue
 
-            # Only form model classes for tables that have a primary key and are not association
+            # Form model classes for tables that have a primary key and are not association
             # tables
             if not table.primary_key:
-                models_by_table_name[table.name] = Model(table)
-            else:
-                model = ModelClass(table)
-                models_by_table_name[table.name] = model
+                first_col_name = table.c.values()[0].name
+                pk = PrimaryKeyConstraint('forced_pk', first_col_name)
+                table.append_constraint(pk)
+            model = ModelClass(table)
+            models_by_table_name[table.name] = model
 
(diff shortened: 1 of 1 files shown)


agronholm:
I'm not inclines to merge this since views can be used as regular tables without issues.

agronholm:
To put it in another way, I am not happy with the generated model deviating from the underlying schema.

KhirilovA:
I just thought that it could be helpful for people working with orm, that won't change the schema itself but will provide a user with a model with a forced pk on first column, so he won't do it by himself

agronholm:
People working with the ORM can still use ordinary Tables in queries.

KhirilovA:
Yes, you can work with queries, but it is much more easier to work with the model than with the table. 
In case you use it nor for a queries, but for some other purposes: like working with columns ;)

Screenshot_1
Screenshot_2
Screenshot_3
.

agronholm:
I'm sorry, but I am not willing to accept this into the project. It's too much of a foot-gun. You'll have to maintain a fork if you want to use this yourself.

KhirilovA:
The last question..
Maybe it would be better if i add an ability to choose on what column to force a pk, by a name or by serial number?)

agronholm:
Nope, sorry, that doesn't help.

PR #190, opened 2022-04-07
Feature/force classes

sschrijver-pon:
Add forceclasses parameter if you only want Python classes instead of `Table(...)`


agronholm:
The ORM cannot use classes w/o primary keys, and association tables as classes don't work with `secondary=` either. I will not accept a PR trying to generate invalid code, unless you can make a very strong case for allowing this.

PR #197, opened 2022-04-24
Add constraint naming convention configuration option

### LINKED ISSUE #193: Add constraint naming convention configuration option
Moving the discussion here from #120

---

> One feature I think would be great to have is to allow customization of constraint naming conventions. We already have `utils.uses_default_name`, but it's currently not super useful because `sqlalchemy.MetaData.naming_convention` is always `{'ix': 'ix_%(column_0_label)s'}` by default. (Reference: https://docs.sqlalchemy.org/en/14/core/constraints.html#the-default-naming-convention)
> 
> We can pass the naming conventions as command line args like
> 
> ```shell
> sqlacodegen --conventions '"ix": "ix_%(column_0_label)s", "uq": "uq_%(table_name)s_%(column_0_name)s", "pk": "pk_%(table_name)s"'
> ```
> 
> or maybe have it read from a JSON file for easier formatting.
> 
> The implementation would be pretty straightforward. We just need to modify the MetaData in `cli.main`.
> 
> We will also need to make `utils.uses_default_name` check `MetaData.naming_convention` properly; right now it just checks for abbreviations: `"fk", "pk", "ix", "ck", "uq"` but the actual Constraint classes (e.g., `PrimaryKeyConstraint`) are valid as well. (Reference: https://docs.sqlalchemy.org/en/14/core/constraints.html#configuring-a-naming-convention-for-a-metadata-collection)
> 
> EDIT: formatting, typo

_Originally posted by @leonarduschen in https://github.com/agronholm/sqlacodegen/issues/120#issuecomment-1087870235_

---

> @leonarduschen sounds good on the general level. I do question naming conventions being passed on the command line. That rabbit hole goes pretty deep when everybody wants to customize every aspect of the code generation step. A YAML or TOML based configuration file would be better. I'm looking forward to seeing what you come up with.
> 
> My intent with the next release is to create a system where you could have your own generator that just gets passed elements to be formatted (tables, classes, whatnot). If would then generate the code, optionally falling back to the default behavior.

_Originally posted by @agronholm in https://github.com/agronholm/sqlacodegen/issues/120#issuecomment-1092788085_

leonarduschen:
Closes #193


agronholm:
FYI, I'm planning one last big refactoring before the final release (one which should resolve the issue of code duplication) and it will probably clash with this PR.

leonarduschen:
Right, I'm actually working on 2 additional issues, details outlined in https://github.com/agronholm/sqlacodegen/issues/193#issuecomment-1109521217

Those issues are only relevant when using these 2 types of tokens:
- `%(referred_column_0_name)s`, `%(referred_column_0_N_name)s`, `%(referred_column_0N_name)s`, etc
- `%(constraint_name)s`

I was thinking of doing everything in one go in this PR, but if you're going to do a big refactoring then perhaps I should just fix those in subsequent PRs? For now we can add a note in the docs that those tokens cannot be used.

Let me know what you think.

agronholm:
I've opened #200 to address the design choice of config file support which we probably need for this.

src/sqlacodegen/cli.py
```diff
@@ -40,6 +52,13 @@ def main() -> None:
     )
     parser.add_argument("--noviews", action="store_true", help="ignore views")
     parser.add_argument("--outfile", help="file to write output to (default: stdout)")
+    parser.add_argument(
+        "--conv",
+        nargs="*",
+        help='constraint naming conventions in "key=template" format \
+e.g., --conv "pk=pk_%%(table_name)s" "uq=uq_%%(table_name)s_%%(column_0_name)s"',
```
leonarduschen, line 59: I added the option to add the naming convention via command line on top of config file (which hasn't been added yet)

I was thinking, for each config option, we can go like this:
- Check if it is available as command line arg
- If not, check if it is available in config file
- If not, fall back to default value

agronholm: How would users know what to enter here, based on this?

leonarduschen: Well we assume users interested in using this option are already familiar with SQLAlchemy's constraint naming convention feature from https://docs.sqlalchemy.org/en/14/core/constraints.html#configuring-constraint-naming-conventions or https://alembic.sqlalchemy.org/en/latest/naming.html. In this case, the `help` string above would enough in my opinion.

We can also add more details in README:
- Link to SQLAlchemy/Alembic docs above
- Maybe also basic info about constraint naming convention and its usage in SQLAlchemy, basically summarizing the docs above
- Things not supported in `sqlacodegen`:
  - Using custom function as naming convention token
  - Using the constraint object name as key instead of mnemonic (e.g., `PrimaryKeyConstraint` cannot be used as key, must use `pk`)

PR #203, opened 2022-04-29
sqlmodel -> sqlmodels

amacfie-tc:


agronholm:
My bad, thanks for the fix!

PR #206, opened 2022-05-06
Restore CITEXT, Geoalchemy2 support

stavvy-rotte:
While experimenting with 3.0.0-rc1 of sqlacodegen I noticed that geoalchemy and sqlalchemy-citext support seemed to be broken.

This PR restores the functionality by importing them if they are available in the cli module where the sqlalchemy metadata is populated via reflection.

This PR also creates a new extra for geoalchemy types and adds some print statements asserting that the libraries loaded correctly (mostly to avoid unused import errors from pyflake, but doubles as confirmation for end-users)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,6 +41,12 @@ considered as tested only under a few environments) specify the ``citext`` extra
     pip install sqlacodegen[citext]
 
 
+To include support for the PostgreSQL ``GEOMETRY``, ``GEOGRAPHY``, and ``RASTER`` types (which should be
+considered as tested only under a few environments) specify the ``geoalchemy2`` extra::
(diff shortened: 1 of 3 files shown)


agronholm:
If I remember correctly, I did intend to support such things but in a better way like as plugins that would be loaded automatically. Adding conditional imports like that was always dirty.

alecglen:
@agronholm Any update on this, or could it possibly get merged in the meantime?

It may be slightly hacky, but I guarantee it's cleaner than the way I've been back-dooring geoalchemy types into my generated classes for the last several months. 😅

agronholm:
I want to include this, but I'm uncomfortable about the lack of tests that make sure this works as intended.

Issue #205, opened 2022-05-04
Option 'use_inflect' raises exception when encountering singular names


Haladesta:
The function `singular_noun()` returns `False` if it gets a name which is already singular (in my case it was the table "sdk").
Because this is not being checked and handled as string, an exception is thrown.

The lines in question are these two:
https://github.com/agronholm/sqlacodegen/blob/7d7a89ab2cfcf9a79c5872c5795cf907eca357e0/src/sqlacodegen/generators.py#L941
https://github.com/agronholm/sqlacodegen/blob/7d7a89ab2cfcf9a79c5872c5795cf907eca357e0/src/sqlacodegen/generators.py#L1005

simply replacing it with something like this should fix it.

    singular_name = self.inflect_engine.singular_noun(preferred_name)
    if isinstance(singular_name, str):
        preferred_name = singular_name

PR #207, opened 2022-05-06
Extend fix for #205 to cover relationships, fix incorrect plural relationship names

stavvy-rotte:
Just squashing two bugs =)

The first one is just #205 but in a new and exciting place, the second is inflect 5.6.0 attempting to make an already plural table name... even more plural

On master:
* test_use_inflect_relationship_singular fails with an attribute error, per #205
* test_use_inflect_relationship_plural fails by generating a relationship with the name 'simple_itemss'

Both issues are fixed by this PR

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -998,13 +998,16 @@ class DeclarativeGenerator(TablesGenerator):
                         preferred_name = column_names[0][:-3]
 
             if "use_inflect" in self.options:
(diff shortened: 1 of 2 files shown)


andrewcarretta:
Code & Tests added in PR https://github.com/agronholm/sqlacodegen/pull/240 will satisfy this PR
- https://github.com/agronholm/sqlacodegen/pull/240

Thanks @stavvy-rotte and @agronholm

PR #211, opened 2022-05-11
Refactor caused citext import to get lost

mzpqnxow:
https://github.com/agronholm/sqlacodegen/pull/125 added seamless support for `citext` when using `psycopg2` and `sqlalchemy`. The import is required to register the type with `psycopg2`. It got dropped during the refactor. This adds it back so users don't have to explicitly import it themselves


agronholm:
Superseded by #206.

PR #228, opened 2022-09-13
Handle inflect not changing a relationship name.

gthank:
BEFORE: If inflect did not change the `preferred_name` for a relationship, the code still overwrote the `preferred_name` (which caused an error when it later triped to invoke `.strip()` on `False`). This broke usage for cases where table names are pluralized, but FK columns are singular, e.g., `users.account_id -> accounts.id`.

AFTER: The relationship naming code uses the pattern from the model naming code to only overwrite the `preferred_name` when inflect hands back a truthy value.

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -1002,9 +1002,13 @@ class DeclarativeGenerator(TablesGenerator):
                     RelationshipType.ONE_TO_MANY,
                     RelationshipType.MANY_TO_MANY,
                 ):
-                    preferred_name = self.inflect_engine.plural_noun(preferred_name)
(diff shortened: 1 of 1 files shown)


andrewcarretta:
Yep, I ran into the same issue. `self.inflect_engine.singular_noun` has the opportunity to return a False. This PR would resolve this issue.

agronholm:
An accompanying test to prevent future regressions is needed to merge this.

andrewcarretta:
Code & Tests added in PR https://github.com/agronholm/sqlacodegen/pull/240 will satisfy this PR
- https://github.com/agronholm/sqlacodegen/pull/240

Thanks @gthank and @agronholm

Issue #236, opened 2022-10-27
Move to sqlalchemy 2.0 beta


PaleNeutron:
I found `sqlalchemy.__all__` does not exist in version 2.0.

rbuffat:
SQLAlchemy 2.0.0 is now released: https://www.sqlalchemy.org/blog/#sqlalchemy-2.0.0-released

Henkhogan:
> I found `sqlalchemy.__all__` does not exist in version 2.0.

https://github.com/agronholm/sqlacodegen/blob/f05d99e2a2a6d04b065ebad4b1d68a69b1e69572/src/sqlacodegen/generators.py#L244

I worked around by changing it to

`elif type_.__name__ in sqlalchemy.__dict__:  # type: ignore[attr-defined]`

amacfie:
Try latest `master`

PR #247, opened 2023-01-30
Pin SQLAlchemy < 2.0.0

amacfie:


amacfie:
Well there's https://github.com/agronholm/sqlacodegen/issues/236

agronholm:
This project doesn't use `sqlalchemy.__all__`.

amacfie:
https://github.com/agronholm/sqlacodegen/blob/f05d99e2a2a6d04b065ebad4b1d68a69b1e69572/src/sqlacodegen/generators.py#L244 ?

agronholm:
Ah...I was looking at the most recent changes where this was eliminated. But indeed on GitHub master, it is still used.

rbuffat:
> Ah...I was looking at the most recent changes where this was eliminated. But indeed on GitHub master, it is still used.

@agronholm Are these changes somewhere available? I would be interested in implementing a generator for SQLAlchemy2 (Although I'm not yet sure how difficult it will be and probably would require some guidance)

agronholm:
I think I still need to make a post-release for the 2.x series since a lot of people are opening issues about problems with sqla 2.0.

Issue #233, opened 2022-09-29
Bad foreign_keys parameter for self-referencing table


amacfie:
DDL in postgres:
```sql
CREATE TABLE document(
  id                    UUID NOT NULL PRIMARY KEY DEFAULT gen_random_uuid(),
  ...
  image_id              UUID
);

ALTER TABLE document ADD CONSTRAINT image_fk
  FOREIGN KEY (image_id)
  REFERENCES document(id) ON DELETE CASCADE;
```

dataclass generated:
```python
@mapper_registry.mapped
@dataclass
class Document:
    __tablename__ = 'document'
    __table_args__ = (
        ForeignKeyConstraint(['image_id'], ['document.id'], ondelete='CASCADE', name='image_fk'),
        PrimaryKeyConstraint('id', name='document_pkey'),
        ...
    )
    __sa_dataclass_metadata_key__ = 'sa'

    id: str = field(metadata={'sa': Column(UUID, server_default=text('gen_random_uuid()'))})
    image_id: Optional[str] = field(default=None, metadata={'sa': Column(UUID)})
    ...

    image: Optional[Document] = field(default=None, metadata={'sa': relationship('Document', remote_side=[id], foreign_keys=[image_id], back_populates='image_reverse')})
    image_reverse: List[Document] = field(default_factory=list, metadata={'sa': relationship('Document', remote_side=[image_id], foreign_keys=[image_id], back_populates='image')})
    ...
```

SQLAlchemy error:
```
sqlalchemy.exc.ArgumentError: Column expression expected for argument 'foreign_keys'; got Field(name='image_id',type='Optional[str]',default=None,default_factory=<dataclasses._MISSING_TYPE object at 0x7ffa9ebd1ab0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({'sa': Column('image_id', UUID(), ForeignKey('document.id'), table=<document>)}),kw_only=False,_field_type=_FIELD).
```

I think the argument should be `"[Document.image_id]"`.

agronholm:
IIRC this is fixed in master. Would you mind trying with that?

amacfie:
I think in my original message I was using master too, but there were other fields and tables.

amacfie:
Issue is gone at least for declarative mode with SQLAlchemy 2

simkessy:
@agronholm I generated my models using version 3.0.0rc2 which I believe has the fix in for this issue but i'm getting and error with this model

```
[...]

PR #249, opened 2023-02-05
Add support for SQLAlchemy 2.0

rbuffat:
The aim of this PR is to add support for SQLAlchemy 2.0 Declarative Mapping. However, I'm neither familiar with the codebase of sqlacodegen nor the new SQLAlchemy 2.0 style. Thus, it might take time until this PR will be finished (if at all). 

Thus, this PR is primarily is mainly created to make this work public for other people that want to contribute or build upon it.


src/sqlacodegen/generators.py
```diff
@@ -241,10 +243,9 @@ def add_import(self, obj: Any) -> None:
 
             if type_.__name__ in dialect_pkg.__all__:
                 pkgname = dialect_pkgname
-        elif type_.__name__ in sqlalchemy.__all__:  # type: ignore[attr-defined]
```
amacfie-tc, line 244: https://github.com/agronholm/sqlacodegen/issues/236#issuecomment-1419178856

rbuffat:
@agronholm If you have time, it would be great if you could have a look at this PR. 

The current state is that it implements the declarative, table and dataclasses generator for sqlachemy2.  However, I need to test the generated models yet with a real database. The sqlmodel generator looks a bit incomplete (e.g. manytomany Relationships). Also sqlmodel seems to be not compatible with Sqlalchemy 2.0. Thus,  I would rather not spend time on it.

There are some known issues: 
Currently, the mapped_column includes the Column type. This would not be necessary. However, I'm not yet sure how well this works with type checker. 

I marked some tests to require Python 3.9, as with this version standard collections can be used as type hints. Theoretically, adding `from __future__ import annotations` should allow this for older Python versions, but then  validate_code() fails. Not yet sure why.

amacfie:
I tested on [this](https://github.com/agronholm/sqlacodegen/issues/233#issuecomment-1263726893) (with the dataclasses generator) and that issue seems to be fixed. (So I'm doubly looking forward to this being merged.) The output was this:
```python
from typing import Optional

from sqlalchemy import ForeignKeyConstraint, PrimaryKeyConstraint, Uuid, text
from sqlalchemy.orm import DeclarativeBase, Mapped, MappedAsDataclass, mapped_column, relationship
import uuid

class Base(MappedAsDataclass, DeclarativeBase):
    pass


class Document(Base):
    __tablename__ = 'document'
    __table_args__ = (
        ForeignKeyConstraint(['image_id'], ['document.id'], ondelete='CASCADE', name='image_fk'),
        PrimaryKeyConstraint('id', name='document_pkey')
    )

    id: Mapped[uuid.UUID] = mapped_column(Uuid, primary_key=True, server_default=text('gen_random_uuid()'))
    image_id: Mapped[Optional[uuid.UUID]] = mapped_column(Uuid)

    image: Mapped['Document'] = relationship('Document', remote_side=[id], back_populates='image_reverse')
    image_reverse: Mapped[list['Document']] = relationship('Document', remote_side=[image_id], back_populates='image')
```
A nitpick, but `import uuid` should be in the previous paragraph. Also, I'm curious if we need the `PrimaryKeyConstraint` since there's `primary_key=True`? I prefer `Foo | None` to `Optional[Foo]` but that's not a big deal.

agronholm:
I just noticed the change from `list` to `List`, but with `from __future__ import annotations`, you can keep using `list`, yes?

rbuffat:
@agronholm I first tried this approach but the test fails in `exec()` of `validate_code()`. See this branch: https://github.com/rbuffat/sqlacodegen/tree/try_annotations and [the](https://github.com/rbuffat/sqlacodegen/actions/runs/4324233982):

Error:
```
The above exception was the direct cause of the following exception:
tests/test_generator_declarative2.py:631: in test_manytomany_nobidi
    validate_code(
tests/conftest.py:16: in validate_code
    exec(generated_code, {})
<string>:10: in <module>
    ???
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/decl_api.py:838: in __init_subclass__
    _as_declarative(cls._sa_registry, cls, cls.__dict__)
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/decl_base.py:248: in _as_declarative
    return _MapperConfig.setup_mapping(registry, cls, dict_, None, {})
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/decl_base.py:329: in setup_mapping
    return _ClassScanMapperConfig(
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/decl_base.py:563: in __init__
    self._scan_attributes()
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/decl_base.py:1005: in _scan_attributes
    collected_annotation = self._collect_annotation(
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/decl_base.py:1233: in _collect_annotation
    extracted = _extract_mapped_subtype(
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/sqlalchemy/orm/util.py:2277: in _extract_mapped_subtype
    raise sa_exc.ArgumentError(
E   sqlalchemy.exc.ArgumentError: Could not interpret annotation Mapped[int].  Check that it uses names that are correctly imported at the module level. See chained stack trace for more hints.
``` 

The Sqlalchemy 1.4 code also uses `typing.List`.  I assume it should be possible to upgrade the generated code using pyupgrade to the required Python version. Thus it should be ok to generate code in the style of the oldest by the project supported Python version.

agronholm:
You need to merge from `master`, resolve the conflicts and fix the `pre-commit` errors.

tests/conftest.py
```diff
@@ -0,0 +1,56 @@
+from textwrap import dedent
+
+import pytest
+from _pytest.fixtures import FixtureRequest
```
agronholm, line 4:
```suggestion
from pytest import FixtureRequest
```

pyproject.toml
```diff
@@ -44,6 +45,8 @@ test = [
     "pytest-cov",
     "psycopg2-binary",
     "mysql-connector-python",
+]
```
agronholm, line 48:
```suggestion
    "mysql-connector-python",
    "sqlacodegen[sqlmodel]",
]
```

agronholm: This will cause the sqlmodel tests to actually run, and restore the lost coverage.

rbuffat: @agronholm [ sqlmodel is not compatible with sqlalchemy 2.0](https://github.com/tiangolo/sqlmodel/blob/main/pyproject.toml#L34), this makes it tricky.

agronholm: Oh right. Hmm, those tests need to be run _somehow_.

agronholm: I think you had the right idea, but couldn't you just have compared the sqlalchemy version to the string `1.4.*` and conditionally installed sqlmodel then?

rbuffat:
> You need to merge from `master`, resolve the conflicts and fix the `pre-commit` errors.

The pre-commit check seemed broken, this I didn't pay too much attention. It still seems to return different errors compared to what is in master, thus not sure how stable these checks are.

pyproject.toml
rbuffat: There are many roads that lead to Rome. This way has the advantage that if one wants to pin SQLalchemy to a certain version only the install line needs to be changed.

agronholm: I'm not going to want to pin it to a specific version.

agronholm:
I updated the pre-commit modules this Monday in master.

.github/workflows/test.yml
```diff
@@ -10,37 +10,42 @@ jobs:
...
-    - name: Test with pytest
-      run: coverage run -m pytest
-    - name: Upload Coverage
-      run: coveralls --service=github
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        COVERALLS_FLAG_NAME: ${{ matrix.test-name }}
-        COVERALLS_PARALLEL: true
-        SQLALCHEMY_WARN_20: "true"
+      - uses: actions/checkout@v3
```
agronholm, line 17: I only just noticed: what caused the reformatting? That makes it hard to see the actual changes.

agronholm: If you could undo the formatting changes and only leave the actual changes in, then I'm ready to merge.

pyproject.toml
```diff
@@ -75,7 +78,8 @@ strict = true
 plugins = ["sqlalchemy.ext.mypy.plugin"]
 
 [tool.pytest.ini_options]
-addopts = "-rsx --tb=short"
+pythonpath = "src/sqlacodegen"
```
agronholm, line 81: This is unnecessary.

pyproject.toml
```diff
@@ -75,7 +78,8 @@ strict = true
...
+addopts = "-rsx --tb=short --import-mode=importlib"
```
agronholm, line 82: Why `--import-mode=importlib`?

agronholm:
I will soon resume work on this project, and would like to incorporate this PR ASAP. I have just a few questions/comments (above).

rbuffat:
@agronholm I found some bugs, e.g. determining the type of collections.  Moreover, after reflecting for some time, I think a major refactoring is needed to make the different generators independent, as currently, the code is getting unnecessarily complex. 

I assume I will have time in the coming weeks to work more on the code.

tomeszmh:
Any news on this one? 

Thanks

agronholm:
Not from my side. This project is pretty low on my priority list.

mhauru:
Hello @rbuffat and @agronholm! Our [project](https://github.com/alan-turing-institute/sqlsynthgen) depends on sqlacodegen (thanks for your work on it!) and we would like to upgrate to SQLAlchemy 2 for various reasons. Judging from the conversation above, this PR seems close to being mergeable. We could put some time into helping get this over the line, but for that it would be very helpful if you could summarise what is still needed. I see some comments by @agronholm above that seem relatively straight-forward, but @rbuffat also mentioned some bugs that they found. Is there a description of them somewhere? Also, there was a mention of a major refactor, that sounds to me like it would be a separate PR and not necessary to finish this, is that correct?

agronholm:
I want this done too, but I've been busy with other projects. I still have one more project that I need to complete the next milestone on before I switch my attention to this one. As such, I welcome any efforts on this PR, but I can't devote any time on it just yet.

pyproject.toml
mhauru: This seems resolved to me since https://github.com/agronholm/sqlacodegen/pull/249/commits/bbfacf7eca7b7a13309a127d672fe885af8686a8, see lines 24-29

mhauru:
I made a commit addressing the comments you left @agronholm, see #275.

I think that covers all the explicit changes you requested. I'm a bit bothered by @rbuffat mentioning he found some bugs, but the tests don't pick them up and if @rbuffat isn't around to tell us more, I'm not sure what to do about them.

pyproject.toml
mhauru: Removed in #275

pyproject.toml
mhauru: Removed in #275

.github/workflows/test.yml
mhauru: Formatting removed in #275

pyproject.toml
rbuffat: pytest recommends to use importlib: https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html

rbuffat:
@mhauru As far as I remember, there are issues with name clashes between column names and imports. E.g. a column named `uuid`. 

```
import uuid
from sqlalchemy import Uuid

class Test(Base):
   uuid: Mapped[uuid.UUID] = mapped_column(Uuid)
```
Additionally, there are issues with identifying the correct types  e.g. for `sqlalchemy.dialects.postgresql.JSONB`

But the biggest issue is the current architecture.  The different generators inherit from each other. With the current changes to support sqlalchemy 2.0 this increases the number of possible paths that can be taken and makes everything in my opinion unmaintainable.  I would suggest making the generator not inheriting from each other and moving common functionality (e.g. determining Python types, field naming, importing of libraries out of the individual generators.

agronholm:
Superseded by #275.

Issue #259, opened 2023-03-20
Enable `--options` to be used more than once


jensenbox:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

The documentation indicates that you can use `--options` multiple times but it appears that you cannot.

For example `--option nojoined --option use_inflect` only gets the last option.

While I understand this is not a feature request it was the only issue type I could create.

### Use case

`sqlacodegen --generator sqlmodels --option nojoined --option use_inflect --outfile njmodels.py postgresql://user:pass@0.0.0.0:5432/mydb`

agronholm:
Which version of sqlacodegen did you try?

jensenbox:
```sqlacodegen --version
3.0.0rc2
```

agronholm:
I've fixed the bug report form so that new such issues can be created.
Are you sure that multiple options cannot be passed? [This line](https://github.com/agronholm/sqlacodegen/blob/master/src/sqlacodegen/cli.py#L24) here indicates otherwise. You tried reversing the order and only the last is applied...?

jensenbox:
I did try reversing the order :) that is exactly how I deduced this issue. I am going to dive a bit deeper into the exact order of what is going on but my current hunch is that one conditional is shadowing the other. I have PyCharm and debugging wired up to the project now and will take another stab in the morning. I will report my findings.

jensenbox:
I plopped in a PR for a documentation update on this: #261 - edit at will. It gets the idea across.

PR #261, opened 2023-03-21
References #259 Update README.rst to include `--options` usage.

jensenbox:
Update docs to reflect how to use `--options` and refer to where this requirement comes from.


agronholm:
I will change the options to be comma delimited like all the other multi-value arguments, for consistency.

PR #262, opened 2023-03-22
Double pluralization of names

### LINKED ISSUE #260: Stop double pluralization of names
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

Most of the tables I am working with come from ActiveRecord where the relationship names are pluralized. In the current implementation a plural_noun of the already plural name like `users` ends up with a name like `userss` instead of `users`.

The easy fix is to first convert it to a singular name then pluralize like:
```
                    singular_inflected_name = self.inflect_engine.singular_noun(preferred_name)
                    inflected_name = self.inflect_engine.plural_noun(singular_inflected_name)
```

This came from around line 1000 in the generators.py

An option for "singularize" or perhaps to not use inflect on relationship names could work too. In my case first converting to singular worked well.

### Use case

When you are generating models from ActiveRecord implementations.

dkratzert:
Fixes #260 Stop double pluralization of names

* The proposal in #260 had the problem that inflect_engine.singular_noun() can be False, e.g. for inflect_engine.singular_noun('version')


agronholm:
Regarding the first test, `simple_item = relationship('SimpleItem', back_populates='container')` is not correct because the related foreign key column does not have a unique constraint on it.

NixBiks:
@dkratzert are you still planning to get this merged?

dkratzert:
Sorry for the delay! I hope to have time for it soon.

sheinbergon:
@dkratzert we are now ready to deal with this issue. We don't one or two tests to cover this change, (including the FK case @agronholm commented about). Can you re-add them?

agronholm:
> @dkratzert we are now ready to deal with this issue. We need one or two tests to cover this change, (including the FK case @agronholm commented about). Can you re-add them?

We could just resurrect and correct the tests from the commit history.

dkratzert:
Please use the tests from the history. I still don't find the time...

sheinbergon (approved):

PR #289, opened 2023-09-28
Update generators.py

Dsolnik:
Allow for spaces in a constraint for enums:

For example, if your constraints was `CHECK IN('a', 'b', 'c')`, this would not be recognized.

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -68,9 +68,9 @@ else:
     from importlib.metadata import version
 
 _sqla_version = tuple(int(x) for x in version("sqlalchemy").split(".")[:2])
-_re_boolean_check_constraint = re.compile(r"(?:.*?\.)?(.*?) IN \(0, 1\)")
+_re_boolean_check_constraint = re.compile(r"(?:.*?\.)?(.*?)\s+IN\s*\(0, 1\)")   
 _re_column_name = re.compile(r'(?:(["`]?).*\1\.)?(["`]?)(.*)\2')
-_re_enum_check_constraint = re.compile(r"(?:.*?\.)?(.*?) IN \((.+)\)")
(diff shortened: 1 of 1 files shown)


agronholm:
I don't think the changes you made do what you think they do.

agronholm:
Basically all your changes do is allow any amount of whitespace before and after the `IN`.

Dsolnik:
Tired Thursday apologies.

Yes, this PR allows for multiple spaces before the check keyword and any amount of white space (including 0) after the check keyword but before the in keyword.

agronholm:
Just what kind of CHECK constraint did you have a problem with that wouldn't work with the existing code? I have a hard time understanding.

Dsolnik:
My constraint was “CHECK IN('a', 'b', 'c')”, notice the lack of spaces after the IN

agronholm:
I thought the RDBMS would normalize these. What RDBMS was this on?

Dsolnik:
Admittedly, this was an oversight when I uploaded the schema but it was
recognized as valid sql (and worked as a constraint) in SQLite and SQL
Server.

agronholm:
Ah, that answers my question. I guess it's only PostgreSQL that normalizes those then.

agronholm:
I will need an accompanying test to ensure that the PR fixes what it says it fixes (and won't break in the future). Meanwhile I'll experiment with sqlite (I don't know how to operate SQL Server).

Dsolnik:
Yea, not only that but SQLite returns back the comment associated with the
constraint (rip).

I didn’t check if this was the case for sql server.

cameronclaero:
@Dsolnik would you also be able to add a fix for https://github.com/agronholm/sqlacodegen/issues/291? In this instance the spaces were are the start and end of the constraint eg " CAP_RESTRICT_LEVEL IN ('NONE', 'WARN', 'RESTRICT') ".  I was in the process of creating a pr but looks like you are already making changes ;)

PR #296, opened 2023-10-17
enable finer selection on tables

sdementen:
allow to express the tables to process via regexp both to explicitly keep tables or to explicitly exclude tables

### DIFF
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+import re
 import sys
 from contextlib import ExitStack
 from typing import TextIO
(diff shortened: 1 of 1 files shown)


agronholm:
This lacks:

- tests
- documentation updates
- an added changelog entry

When these have been added, I will review this PR.

sdementen:
I do this with pleasure if you think the feature is worth. Is it?
And is there any test already that tests the --table argument? I wanted to
add a test but could not find one to extend/adapt

agronholm:
There are no tests for it that I can see. It would be nice to test both ways. Do you mind adding those too?

sdementen:
If you could provide (even in comment in this thread) a skeleton of test
for the current --table feature, it would be very useful. Designing a test
with a fake database + fake CLI call with arguments is too challenging for
me. However, once I have a running test, then I will be able to duplicate
it and enrich it to cover the 'ew feature.

Issue #302, opened 2023-11-16
SQLModel codegen produces syntax errors


NixBiks:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Sqlacodegen version

3.0.0rc3

### SQLAlchemy version

2.0.23

### RDBMS vendor

PostgreSQL

### What happened?

The screenshot shows some of the issues.

1. There are unused imports (can easily be fixed by running ruff after codegen)
2. Missing imports, i.e. `mapped_column`.
3. Syntax error due to a `,` in line 22.

<img width="1001" alt="Screenshot 2023-11-16 at 09 18 05" src="https://github.com/agronholm/sqlacodegen/assets/16841792/0512f107-0c51-4de9-878d-e86014d715b3">

### Database schema for reproducing the bug

```sql
CREATE TABLE
  logs (
    timestamp timestamp NOT NULL,
    level text NOT NULL,
    message text NOT NULL,
    schedule_id bigint REFERENCES schedules (id),
    worker_id text,
    run_id text,
    id bigint GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY
  );

-- sort by timestamp
CREATE INDEX logs_timestamp_desc_idx ON logs (timestamp DESC);

-- create view of logs
CREATE VIEW logs_view AS
  SELECT
    timestamp,
    level,
    message,
    schedule_id,
    worker_id,
    run_id
  FROM logs
  ORDER BY timestamp DESC;
```

madsenwattiq:
I see this as well.  In order to fix broken code, I had to manually add the "metadata" variable to all association tables that didn't have primary keys - that's the bug with the double commas.  After that, it works in 3.0.0rc3.

agronholm:
It's not supposed to use `mapped_column` at all. All the tests expect that it uses `Column()` and not `mapped_column()`. The latter is a thing only on SQLAlchemy 2. Before `sqlmodel` 0.0.12 (released 7 hours ago), it didn't even work with SQLAlchemy 2 at all.

michelmetran:
> I'm pretty sure that this was caused by `sqlmodel now supporting SQLAlchemy 2, and sqlacodegen wasn't ready for that.

Thank you for the information!

After that, I used an environment with SQLAlchemy < 2 and it worked!

```
[...]

PR #306, opened 2023-12-01
Fix Column and metadata in SQLModel Generator

THUzxj:
I encountered a similar problem with #302 and I tried to fix it, including

- Force to use `Column` instead of `mapped_column`, through passing `is_table=True`
- Modify `generate_base` to generate `metadata` variable if `Model` exists

Maybe we can also refactor `generate_models` to decouple the function of getting `models` and collecting imports.

It is probably just a temporary solution to the issue. Thus it's up to you whether to merge. Thanks.

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -1528,6 +1528,84 @@ class SQLModelGenerator(DeclarativeGenerator):
             base_class_name=base_class_name,
         )
 
+    def generate_models(self) -> list[Model]:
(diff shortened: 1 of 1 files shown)


agronholm:
I am working on a proper fix for the problem. Most of the time, `Column` should not be used by the generator.

sheinbergon:
@agronholm  @THUzxj  @MRigal  the fix for this  issue has been merged in#358. Closeing

PR #322, opened 2024-04-13
add: support for pydantic models - first cut

mavwolverine:


agronholm:
Is this not basically covered by SQLModel support?

Issue #329, opened 2024-06-04
Issue when generating Postgres DOMAIN type


yb-leakmited:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Sqlacodegen version

3.0.0rc5

### SQLAlchemy version

2.0.30

### RDBMS vendor

PostgreSQL

### What happened?

I used `sqlacodegen --generator tables` to generate my SQLAlchemy tables. But it seems that for Domain it is not working properly.
I have a Domain type that is the following : 

![image](https://github.com/agronholm/sqlacodegen/assets/122872636/77150784-3bb8-4b0b-8369-923a5926ccfa)

sqlacodegen generates the following line:

```SQL
....
Column('lang', DOMAIN('language_tag_type', TEXT(), constraint_name='valid_language_tag', not_null=False, check= < sqlalchemy.sql.elements.TextClause object at 0x000002173F60A1D0 >), nullable=False),
....
```
The check string is not correctly generated. If think It just needs to call to_string() instead of giving the TextClause object no ?

Related closed PR on SQLAlchemy : https://github.com/sqlalchemy/sqlalchemy/commit/017fd9ae0645eaf2a0fbdd067d10c721505b018c

Note : I started python & SQLAchemy not long ago
Note 2 : I'll try to create a repo to reproduce this bug
Note 3 : THANK YOU FOR CREATING THIS TOOL 😄 , it is simple and amazing, hopefully one day it will be SQLAlchemy and maintained by it 

### Database schema for reproducing the bug

```SQL
CREATE DOMAIN public.language_tag_type
    AS text;

ALTER DOMAIN public.language_tag_type
    ADD CONSTRAINT valid_language_tag CHECK (VALUE ~ '^[a-zA-Z]{1,8}(-[a-zA-Z0-9]{1,8})*$'::text);

CREATE TABLE IF NOT EXISTS public."user"
(
    id uuid NOT NULL DEFAULT 'gen_random_uuid()',
    lang language_tag_type COLLATE pg_catalog."default" NOT NULL,
    CONSTRAINT user_pkey PRIMARY KEY (id),
)

```

amacfie-tc:
And the generated Python type is `Any` which SQLAlchemy doesn't like

agronholm:
I don't think DOMAIN types can be reflected by SQLAlchemy, at this point anyway.

amacfie:
[...]

PR #338, opened 2024-07-03
Handle TextClause objects in DOMAIN expressions

amacfie:
Related to #329

The sqlalchemy dependency version was changed to get [this PR](https://github.com/sqlalchemy/sqlalchemy/pull/10729).

The [DOMAIN class](https://github.com/sqlalchemy/sqlalchemy/blob/c8d73b74733fa9f12b7b95a5885e808e6b31f1c6/lib/sqlalchemy/dialects/postgresql/named_types.py#L384) has
* two parameters that can take a `TextClause`: these get wrapped with `text()` as needed, similarly to `server_default` params
* one parameter which is a SQLAlchemy type: this gets imported, similarly to array item types

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -38,7 +38,7 @@ from sqlalchemy import (
     Text,
     UniqueConstraint,
 )
-from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import DOMAIN, JSONB
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.exc import CompileError
 from sqlalchemy.sql.elements import TextClause
@@ -222,6 +222,13 @@ class TablesGenerator(CodeGenerator):
                 or column.type.astext_type.length is not None
             ):
                 self.add_import(column.type.astext_type)
+        elif isinstance(column.type, DOMAIN):
+            self.add_import(column.type.data_type.__class__)
+            if (
+                isinstance(column.type.default, TextClause)
+                or isinstance(column.type.check, TextClause)
+            ):
+                self.add_literal_import("sqlalchemy", "text")
 
         if column.default:
             self.add_import(column.default)
@@ -533,6 +540,12 @@ class TablesGenerator(CodeGenerator):
                 and coltype.astext_type.length is None
             ):
                 del kwargs["astext_type"]
+
+        if isinstance(coltype, DOMAIN):
+            if isinstance(coltype.default, TextClause):
+                kwargs["default"] = render_callable("text", repr(coltype.default.text))
+            if isinstance(coltype.check, TextClause):
+                kwargs["check"] = render_callable("text", repr(coltype.check.text))
 
         if args or kwargs:
             return render_callable(coltype.__class__.__name__, *args, kwargs=kwargs)
--- a/tests/test_generator_tables.py
+++ b/tests/test_generator_tables.py
@@ -189,6 +189,41 @@ def test_enum_detection(generator: CodeGenerator) -> None:
 
 
 @pytest.mark.parametrize("engine", ["postgresql"], indirect=["engine"])
+def test_domain(generator: CodeGenerator) -> None:
+    Table(
+        "simple_items",
+        generator.metadata,
+        Column(
+            'postal_code',
+            postgresql.DOMAIN(
+                'us_postal_code',
+                Text,
+                constraint_name='valid_us_postal_code',
+                not_null=False,
+                check=text("VALUE ~ '^\\d{5}$' OR VALUE ~ '^\\d{5}-\\d{4}$'"),
+            ),
+            nullable=False,
+        ),
+    )
+
+    validate_code(
+        generator.generate(),
+        """\
+        from sqlalchemy import Column, MetaData, Table, Text, text
+        from sqlalchemy.dialects.postgresql import DOMAIN
+
+        metadata = MetaData()
+
+
+        t_simple_items = Table(
+            'simple_items', metadata,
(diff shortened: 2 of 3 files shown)


src/sqlacodegen/generators.py
```diff
@@ -547,6 +553,12 @@ def render_column_type(self, coltype: object) -> str:
             ):
                 del kwargs["astext_type"]
 
+        if isinstance(coltype, DOMAIN):
+            if isinstance(coltype.default, TextClause):
+                kwargs["default"] = render_callable("text", repr(coltype.default.text))
+            if isinstance(coltype.check, TextClause):
+                kwargs["check"] = render_callable("text", repr(coltype.check.text))
```
agronholm, line 560: I'm wondering whether this should be a generic thing – turning any `TextClause` instance into `text()` for any argument passed to a callable?

sheinbergon (approved):
LGTM

src/sqlacodegen/generators.py
agronholm: To put it differently, is there _any_ case where we would want to render something like `< sqlalchemy.sql.elements.TextClause object at 0x000002173F60A1D0 >`? Because I don't think so, and therefore I would like to convert _all_ kwargs in this manner, thus eliminating the need to special case `DOMAIN`.

agronholm: Don't we always add the import for any column type we render anyway?

agronholm: Oh, my bad, it was `column.type.data_type`. Yeah, we need that one. I wonder if there's a way to get this automatically. But for now I'm fine with special casing `DOMAIN`.

agronholm: So the only change to be made is to generalize `TextClause` -> `text()`.

sheinbergon: @andrew222651 would you want to take care of it, or would you want me to do that?

tests/test_generator_tables.py
```diff
@@ -237,9 +311,9 @@ def test_mysql_column_types(generator: CodeGenerator) -> None:
         "simple_items",
         generator.metadata,
         Column("id", mysql.INTEGER),
-        Column("name", mysql.VARCHAR(255)),
-        Column("double", mysql.DOUBLE(1, 2)),
-        Column("set", mysql.SET("one", "two")),
+        Column("name", mysql.VARCHAR(255)),  # type: ignore[no-untyped-call]
```
sheinbergon, line 314: @agronholm are thse new ruff / pep8 standards we need to exclude?

sheinbergon: Why do we suddenly need these mypy exclusions?

agronholm: I thought you asked about Ruff / PEP 8 standards. But yeah, mypy keeps changing and sometimes it triggers errors it didn't before. And also this PR changes the minimum mypy version and removes the `mypy` extra, so maybe that's why?

sheinbergon (changes requested):
2 more fixes required. Please see open discussions

tests/test_generator_tables.py
sheinbergon: Ah, yes I see it here https://docs.sqlalchemy.org/en/20/orm/extensions/mypy.html

So, should we tell pre-commit to exclude this test for the time being?

agronholm: Why? I didn't think the impact was that bad – a few type ignore comments? We can also ignore this specific error category altogether if necessary.

sheinbergon: Yes, I prefer we do that. Otherwise, it will create inconsistencies in the test where some will have these annotations and others won't.

sheinbergon:
@andrew222651 thank you. I want to have a look at your changes carefully before moving forward

agronholm:
These changes address the mypy issue but not the `TextClause` generalization.

agronholm:
Something clearly went wrong with that. Did the test pass for you locally?

andrew222651:
Tests passing now, let me know if you want me to implement it differently.

src/sqlacodegen/generators.py
```diff
@@ -514,12 +520,16 @@ def render_column_type(self, coltype: object) -> str:
 
             value = getattr(coltype, param.name, missing)
             default = defaults.get(param.name, missing)
+            if isinstance(value, TextClause):
+                rendered_value = render_callable("text", repr(value.text))
+            else:
+                rendered_value = repr(value)
```
agronholm, line 526: Add a blank line to distinguish between the two unrelated if...else blocks.
```suggestion

                rendered_value = repr(value)
```

src/sqlacodegen/generators.py
```diff
@@ -514,12 +520,16 @@ def render_column_type(self, coltype: object) -> str:
...
             if value is missing or value == default:
```
agronholm, line 524: Add a blank line to distinguish between the two unrelated if...else blocks.
```suggestion

            if value is missing or value == default:
```

agronholm (approved):

agronholm:
> @andrew222651 more of a general question. What's the reason for adding type `cast` in two places?

I assume this was to fix mypy errors.

sheinbergon:
> > @andrew222651 more of a general question. What's the reason for adding type `cast` in two places?
> 
> I assume this was to fix mypy errors.

That's my guess as well. I reverted one of them, and indeed pre-commit fails with mypy errors. However, there are other locations with type compat issues that just don't trigger mypy errors. In any case, I answered my own questions. LGTM

sheinbergon:
@agronholm we forgot to update `CHANGES.rst`

agronholm:
Ah. I was so focused on the code changes that I forgot about this. Shall I add a PR template? I have them in most other projects I actively work on.

sheinbergon:
> Ah. I was so focused on the code changes that I forgot about this. Shall I add a PR template? I have them in most other projects I actively work on.

Sounds good

PR #349, opened 2024-11-21
Support sequences on TimescaleDB like on PostgreSQL

MRigal:
Tiny fix, as else sqlacodegen is also working with TimescaleDB and timeseries hypertables

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -72,7 +72,7 @@ _re_column_name = re.compile(r'(?:(["`]?).*\1\.)?(["`]?)(.*)\2')
 _re_enum_check_constraint = re.compile(r"(?:.*?\.)?(.*?) IN \((.+)\)")
 _re_enum_item = re.compile(r"'(.*?)(?<!\\)'")
 _re_invalid_identifier = re.compile(r"(?u)\W")
-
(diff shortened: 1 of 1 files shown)


agronholm:
Wait, what? There's a specific dialect for timescaledb despite it just being a postgresql extension?

MRigal:
Not a real official dialect, but using this extension: https://github.com/dorosch/sqlalchemy-timescaledb gives some nice enhanced support for hypertables

agronholm:
Does it improve the output of `sqlacodegen` when run against a database containing hypertables?

MRigal:
I've just tested, the answer is "no". It probably misses in https://github.com/dorosch/sqlalchemy-timescaledb/blob/master/sqlalchemy_timescaledb/dialect.py additional features that would look for hypertables on some Inspector methods: https://docs.sqlalchemy.org/en/20/core/reflection.html#sqlalchemy.engine.reflection.Inspector

I'm not familiar enough with it to contribute directly, but if you are also interested and point me towards a possible implementation, I could submit a PR for it

agronholm:
Ok, so is there any reason to run sqlacodegen with the timescaledb dialect instead of postgresql then?

MRigal:
No other one than to use the same engine URL as elsewhere in the code

PR #358, opened 2025-01-09
SQLModel Code generation fixes 

sheinbergon:
Fixes [302](https://github.com/agronholm/sqlacodegen/issues/302)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 test = [
+    "sqlacodegen[sqlmodel]",
     "pytest >= 7.4",
     "coverage >= 7",
     "psycopg2-binary",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 test = [
-    "sqlmodel",
     "pytest >= 7.4",
     "coverage >= 7",
     "psycopg2-binary",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,12 +41,13 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 test = [
+    "sqlmodel",
     "pytest >= 7.4",
     "coverage >= 7",
     "psycopg2-binary",
     "mysql-connector-python",
 ]
-sqlmodel = ["sqlmodel >= 0.0.12"]
+sqlmodel = ["sqlmodel >= 0.0.22"]
 citext = ["sqlalchemy-citext >= 1.7.0"]
 geoalchemy2 = ["geoalchemy2 >= 0.11.1"]
 pgvector = ["pgvector >= 0.2.4"]
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -63,8 +63,8 @@ def main() -> None:
 
     if args.generator == "sqlmodels":
         print(
-            "NOTICE: VIEW models will not be generated when using the 'sqlmodels' "
-            "generator"
+            "VIEW models will not be generated when using the 'sqlmodels' generator",
+            file=sys.stderr,
         )
 
     if args.version:
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -56,7 +56,7 @@ def main() -> None:
     parser.add_argument(
         "--noviews",
         action="store_true",
-        help="ignore views (always true for sqlmodel generator)",
+        help="ignore views (always true for sqlmodels generator)",
     )
     parser.add_argument("--outfile", help="file to write output to (default: stdout)")
     args = parser.parse_args()
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -61,12 +61,6 @@ def main() -> None:
     parser.add_argument("--outfile", help="file to write output to (default: stdout)")
     args = parser.parse_args()
 
-    if args.generator == "sqlmodels":
-        print(
-            "VIEW models will not be generated when using the 'sqlmodels' generator",
-            file=sys.stderr,
-        )
-
     if args.version:
         print(version("sqlacodegen"))
         return
@@ -91,14 +85,22 @@ def main() -> None:
     tables = args.tables.split(",") if args.tables else None
     schemas = args.schemas.split(",") if args.schemas else [None]
     options = set(args.options.split(",")) if args.options else set()
-    for schema in schemas:
-        metadata.reflect(
-            engine, schema, (args.generator != "sqlmodels" and not args.noviews), tables
-        )
 
     # Instantiate the generator
     generator_class = generators[args.generator].load()
     generator = generator_class(metadata, engine, options)
+
+    if not generator.views_supported:
+        name = generator_class.__name__
+        print(
+            f"VIEW models will not be generated when using the '{name}' generator",
+            file=sys.stderr,
+        )
+
+    for schema in schemas:
+        metadata.reflect(
+            engine, schema, (generator.views_supported and not args.noviews), tables
+        )
 
     # Open the target file (if given)
     with ExitStack() as stack:
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -61,6 +61,12 @@ def main() -> None:
     parser.add_argument("--outfile", help="file to write output to (default: stdout)")
     args = parser.parse_args()
 
+    if args.generator == "sqlmodels":
+        print(
+            "NOTICE: VIEW models will not be generated when using the 'sqlmodels' "
+            "generator"
+        )
+
     if args.version:
         print(version("sqlacodegen"))
         return
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -53,7 +53,11 @@ def main() -> None:
     parser.add_argument(
         "--tables", help="tables to process (comma-delimited, default: all)"
     )
-    parser.add_argument("--noviews", action="store_true", help="ignore views")
+    parser.add_argument(
+        "--noviews",
+        action="store_true",
+        help="ignore views (always true for sqlmodel generator)",
+    )
     parser.add_argument("--outfile", help="file to write output to (default: stdout)")
     args = parser.parse_args()
 
@@ -82,7 +86,9 @@ def main() -> None:
     schemas = args.schemas.split(",") if args.schemas else [None]
     options = set(args.options.split(",")) if args.options else set()
     for schema in schemas:
-        metadata.reflect(engine, schema, not args.noviews, tables)
+        metadata.reflect(
+            engine, schema, (args.generator != "sqlmodels" and not args.noviews), tables
+        )
 
     # Instantiate the generator
     generator_class = generators[args.generator].load()
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -134,10 +134,6 @@ class TablesGenerator(CodeGenerator):
         self.imports: dict[str, set[str]] = defaultdict(set)
         self.module_imports: set[str] = set()
 
-    @property
-    def mapped_columns_supported(self) -> bool:
-        return False
-
     def generate_base(self) -> None:
         self.base = Base(
             literal_imports=[LiteralImport("sqlalchemy", "MetaData")],
@@ -486,7 +482,10 @@ class TablesGenerator(CodeGenerator):
         if comment:
             kwargs["comment"] = repr(comment)
 
-        if is_table or not self.mapped_columns_supported:
+        return self.render_column_callable(is_table, *args, **kwargs)
+
+    def render_column_callable(self, is_table: bool, *args: Any, **kwargs: Any) -> str:
+        if is_table:
             self.add_import(Column)
             return render_callable("Column", *args, kwargs=kwargs)
         else:
@@ -732,10 +731,6 @@ class DeclarativeGenerator(TablesGenerator):
         self.base_class_name: str = base_class_name
         self.inflect_engine = inflect.engine()
 
-    @property
-    def mapped_columns_supported(self) -> bool:
-        return True
-
     def generate_base(self) -> None:
         self.base = Base(
             literal_imports=[LiteralImport("sqlalchemy.orm", "DeclarativeBase")],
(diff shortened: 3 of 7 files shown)


sheinbergon:
@agronholm notice this PR also completely removes mapped_column for `SQLModel` , which according to  https://github.com/fastapi/sqlmodel/discussions/970 is still broken. I'm guessing you `SQLModel` broke when you introduced it to the codebase. All tests are now passing  :partying_face: , and I believe this takes care of another open issue as well

sheinbergon:
@agronholm I also tested generated sqlmodel with both orm functionalities (`get`,`get_one`,`add`) as well as well SQL queries (`select(...).where()`),  and both seem to function properly

pyproject.toml
```diff
@@ -41,12 +41,13 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 test = [
+    "sqlmodel",
```
agronholm, line 44:
```suggestion
    "sqlacodegen['sqlamodel']",
```

src/sqlacodegen/generators.py
```diff
@@ -1358,17 +1369,23 @@ def __init__(
...
             declarations=[],
-            metadata_ref="",
+            metadata_ref="metadata",
+            table_metadata_declaration="metadata = MetaData()",
         )
 
     def collect_imports(self, models: Iterable[Model]) -> None:
         super(DeclarativeGenerator, self).collect_imports(models)
+        if any(isinstance(model, Model) for model in models):
+            self.add_literal_import("sqlalchemy", "MetaData")
```
agronholm, line 1387: Separate unrelated control blocks with a blank line for readability.
```suggestion
            self.add_literal_import("sqlalchemy", "MetaData")

```

agronholm: But also, why are we unconditionally importing MetaData, particularly in the case of SQLModels which don't even use it directly?

src/sqlacodegen/generators.py
```diff
@@ -134,6 +134,10 @@ def __init__(
         self.imports: dict[str, set[str]] = defaultdict(set)
         self.module_imports: set[str] = set()
 
+    @property
+    def mapped_columns_supported(self) -> bool:
+        return False
```
agronholm, line 139: I think the individual generator should render its column representations how it sees fit, instead of the base implementation querying this property.

agronholm (changes requested):

pyproject.toml
sheinbergon: It's probably better to just change the CI to install `.[test,sqlmodel]` instead of defining `sqlmodel` dependency twice, so I'm gonna make this change instead

agronholm: Then you'd have to keep making these changes in both `pyproject.toml` every time you add something. I've used this method in my other projects and been happy with it. Where is the duplication here?

src/sqlacodegen/generators.py
sheinbergon: Part of the broken codegen was a Table def being generated for SQL views without metadata. The thing is, imports are being collected / generated before the models/tables are being generated, so trying to resolve that decision is going to introduce more complexity, and if the price was having an unused `MetaData`  definition at the top of the generated python module, I assumed it was a price worth paying. Let me know if you feel this is a blocker for merging this PR.

src/sqlacodegen/generators.py
sheinbergon: Generally speaking, I would have wholeheartedly agreed with you when buidling `sqlacodegen` from the ground up. 

However, looks like you've made class inhertiance tree (each generated builds upon the previous one) a core design principal in this library,  `mapped_columns_supported` is observed in two places, and if you look at the [render_column](https://github.com/agronholm/sqlacodegen/blob/d9e328266d40245d54323358ce785cdcaa975bb0/src/sqlacodegen/generators.py#L399) function, it's a already a 100 lines function, already calling on other instance functions.

IMO, duplicating all of this just to avoid a single call to an overridable function, when everything is already based on inheritance would be unwise, especially as part of this PR. I believe you created another PR to simplify the codebase, something like should be tackled as part such effort.

src/sqlacodegen/generators.py
sheinbergon: Was it always your intention to generate a table represenation for views when using the `sqlmodel` generator

pyproject.toml
sheinbergon: The duplication would be defining `sqlmodel` in both tests and as the separate sqlmodel group. As you already Following the fix, there's no need to maintain a definition in separate files as the CI installes `.[test,sqlmodel]` and a versioned `sqlmodel` dependency is defined in a single place

sheinbergon: Also, corret me if I'm wron, but using `sqlacodegen['sqlmodel']` would result in a remote dependency being resolved, while we're aiming to relate to the local one

agronholm: No, it would not. Installers know what's being installed (`sqlacodegen`), and doing this will simply activate other extras in the same package.

src/sqlacodegen/generators.py
agronholm: But are you proposing this property to be part of the public interface of SQLACodegen going forward? If so, it would constrain any such simplification efforts.

agronholm: I'm also not sure what you think would be duplicated. Certainly I would never want complex rendering functions to be duplicated, but instead allow subclasses to render the abstract model however they see fit.

pyproject.toml
sheinbergon: Ok, fixed according to your spec.

src/sqlacodegen/generators.py
sheinbergon: I'll try and clarify.

`render_column` is a complex 100 lines long function. It's one of the 2 callers to `mapped_columns_supported`, and is responsible for the apsect of colmun rendering.

When you say "I think the individual generator should render its column representations how it sees fit",  I understand that I need to create an individual implementation of `render_column` for each generator,  even though 95 lines out of the function's 100 are identical, and only the if-else block should change, depending on the logical resolve of  `mapped_columns_supported`. 

So, that would cause duplication, which is bad.

```python
        if is_table or not self.mapped_columns_supported:
            self.add_import(Column)
            return render_callable("Column", *args, kwargs=kwargs)
        else:
            return render_callable("mapped_column", *args, kwargs=kwargs)
```

I can move the above entire block to a separate function so instead of `mapped_columns_supported` function / property, and we'll have a function dealing with just the column statement rendering, though IMO, it's just a different way of doing exactly the same thing, and you'll wind with exactly the same amount of functions (because the code is already based on inheritence) . 

If that's what you had in mind, let me know
If you had something else in mind, also let me know :)

As for complexity concerns. A simplification rewrite will require re-thinking of a lot of the generators design principals. IMO, it's more crucial the get a proper stable version release based on the current work, so the community can keep using your project, knowing it's alive and maintained.

agronholm: The block you pasted should probably be abstracted into something like `self.render_column()` to defer the decision on how to render the column in the code.

I will say that the separation of concerns as to what's considered part of the model and what's part of the rendering logic has been the most difficult aspect of the design thus far.

> As for complexity concerns. A simplification rewrite will require re-thinking of a lot of the generators design principals. IMO, it's more crucial the get a proper stable version release based on the current work, so the community can keep using your project, knowing it's alive and maintained.

Don't forget that once there's a stable version out, the maintainers will have to support whatever interface comes with it at least until the next major release. I was quite unhappy with the interface I had thus far, and that's why I had not released the final version.

sheinbergon: Ok, I'll change it. 
 
 However,  for this change to make sense, we should decide if table generation for view should remain supported. Otherwise, this separation will not make any sense.
 
 Also, note that dataclasses and delcarative will share the same implementation, so it's not going to abstract, just overriden on by the SQLModelGenerator

sheinbergon: Fixed. Let me know WYT

agronholm: > However, for this change to make sense, we should decide if table generation for view should be kept (see other comment). Otherwise, this separation will not make any sense.

Hmm, I didn't find any reference whatsoever to views in the SQLModel documentation. If indeed this is unsupported there then we should probably skip views altogether with the `sqlmodel` generator then.

But a much more common issue is the lack of primary keys in tables. I've seen plenty of badly designed schemas that don't have primary keys, and then people try to generate code for them and wonder why classes are not being generated, and even going as far as insisting that we should support this anyway!

src/sqlacodegen/generators.py
sheinbergon: @agronholm so the decision should be to skip views for sqlmodel?

sheinbergon: @agronholm please share your decision

agronholm: If views cannot be represented as SQLModel models, then they should be skipped, but do check first if that is the case.

sheinbergon: @agronholm it's not that they cannot be represented, you can technically, create a model matching a view and query it. However it will be a read-only model, which is not something that technically exists in SQLModel. Views do not really fit ORM CRUD operations paradigm. My suggestion would be to drop view generation for SQLModel. If the user still want view bindings, he can just generate tables alongside the sqlmodels.

That's my call, let me know if you agree, and I'll get to it.

src/sqlacodegen/generators.py
motey: @agronholm 
> But a much more common issue is the lack of primary keys in tables. I've seen plenty of badly designed schemas that don't have primary keys, and then people try to generate code for them and wonder why classes are not being generated, and even going as far as insisting that we should support this anyway!

Yep, had exactly this. One workaround is to define all columns as candidate keys with `__mapper_args__ = {"primary_key": [...`. this way we give sqlmodel/sqlachemy a way to work with the record but also not change the actual database schema. Also its relative easy to generate; just need generate  a list of all column names of the table and attach it to `__mapper_args__ = {"primary_key": [`

https://docs.sqlalchemy.org/en/20/orm/declarative_config.html#mapper-configuration-options-with-declarative

agronholm: > @agronholm
> 
> > But a much more common issue is the lack of primary keys in tables. I've seen plenty of badly designed schemas that don't have primary keys, and then people try to generate code for them and wonder why classes are not being generated, and even going as far as insisting that we should support this anyway!
> 
> Yep, had exactly this. One workaround is to define all columns as candidate keys with `__mapper_args__ = {"primary_key": [...`. this way we give sqlmodel/sqlachemy a way to work with the record but also not change the actual database schema. Also its relative easy to generate; just need generate a list of all column names of the table and attach it to `__mapper_args__ = {"primary_key": [`
> 
> https://docs.sqlalchemy.org/en/20/orm/declarative_config.html#mapper-configuration-options-with-declarative

Faking a primary key just to make the table definitions work with the ORM is dubious at best. Equality tests will be expensive and you won't be able to create duplicate rows. Duplicate rows loaded from the database won't be visible from queries.

motey: yea, its just a possible "workaround" that worked in my case. The "dubious" part are tables without any primary keys from my perspective :D 

> you won't be able to create duplicate rows. Duplicate rows loaded from the database won't be visible from queries.

Good point! that is an issue i did not see/did not have in my usecase.

sheinbergon: @agronholm @motey this is not a part of the scope of this work of this PR, so let's leave this as off topic for now. It'll not deal with lack of primary keys. We can deal with it in a later PR, if the masses request it.
@agronholm I'll will complete the final fixes later this week

src/sqlacodegen/generators.py
agronholm: Yeah, skip views then. But make sure there's at least a warning generated when this is done, so users won't be left thinking this is a bug.

src/sqlacodegen/generators.py
sheinbergon: Done

sheinbergon:
@agronholm all PR comments have been fixed according to your guidance, and conversations resolved. Hoping we can move forwrad

src/sqlacodegen/cli.py
```diff
@@ -53,10 +53,20 @@ def main() -> None:
...
+        help="ignore views (always true for sqlmodels generator)",
+    )
     parser.add_argument("--outfile", help="file to write output to (default: stdout)")
     args = parser.parse_args()
 
+    if args.generator == "sqlmodels":
+        print(
+            "NOTICE: VIEW models will not be generated when using the 'sqlmodels' "
+            "generator"
+        )
```
agronholm, line 68: What I meant by a warning was something that would only be printed to `stderr` if there are views in the schema and the user is using the `sqlmodels` generator. I don't want backend specific code in the cli module anyway if that can be avoided.

agronholm (changes requested):

src/sqlacodegen/cli.py
sheinbergon: `Metadata.reflect(...)` happens here. as part of the CLI. you need to differentiate here, why not add warning here?

agronholm: Why do you need to differentiate here? Why not just skip views in the generator?

sheinbergon: Because you chose to do metadata.reflection here. When the code reaches the table selection a view is just another table. I have no simple way to differenatiate there. THe only reasonible place to differentiate is where `metadata.reflect(...)` is called, which is, with the current state of your codebase, withing the `cli` module.

Changed to print to stderr

agronholm: Hmm...okay, I'll think about this a little bit.

sheinbergon: OK.  Just bear in mind we're not trying to solve all of the modeling concerns with this PR. while I understand you don't want to introduce quirks and smells into the codebase, building upon existing design choices to make progress with and gain small victories, that would eventually bring more confidence and functionality  to your userbase, is a sound choice, for the very least.

agronholm: Ok, how about a compromise: reverse the order of the reflection and generator instantiation, and then ask the generator if it supports views?

sheinbergon: So, add some sort of property to the generator, allowing it to call if it supports view generation or not?

agronholm: I think you need to do something along the lines of:
```py
# Get the generator class
generator_class = generators[args.generator].load()
if not generator_class.supports_views:
    args.noviews = True

...

generator = generator_class(metadata, engine, options)

agronholm: > So, add some sort of property/method to the generator, allowing it to call if it supports view generation or not?

Yes. That would reduce the code small significantly.

sheinbergon: Better now?

agronholm (approved):

agronholm:
Thanks!

PR #359, opened 2025-01-12
Minor doc correction

poulter7:


agronholm:
Thanks.

poulter7:
The dependent package is called sqlmodel, but I think it's better that the doc represents your package's accepted params.

PR #363, opened 2025-01-24
Fix SQLAlchemy enum generation

sheinbergon:
Fixes https://github.com/agronholm/sqlacodegen/issues/361

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
         cache: pip
         cache-dependency-path: pyproject.toml
     - name: Install dependencies
-      run: pip install -e .[test,sqlmodel]
+      run: pip install -e .[test]
     - name: Test with pytest
       run: coverage run -m pytest
     - name: Upload Coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
         cache: pip
         cache-dependency-path: pyproject.toml
     - name: Install dependencies
-      run: pip install -e .[test]
+      run: pip install -e .[test,sqlmodel]
     - name: Test with pytest
       run: coverage run -m pytest
     - name: Upload Coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,13 +21,13 @@ classifiers = [
     "Topic :: Software Development :: Code Generators",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 dependencies = [
     "SQLAlchemy >= 2.0.23",
     "inflect >= 4.0.0",
@@ -44,7 +44,7 @@ test = [
     "sqlacodegen[sqlmodel]",
     "pytest >= 7.4",
     "coverage >= 7",
-    "psycopg2-binary",
+    "psycopg[binary]",
     "mysql-connector-python",
 ]
 sqlmodel = ["sqlmodel >= 0.0.22"]
@@ -66,21 +66,23 @@ version_scheme = "post-release"
 local_scheme = "dirty-tag"
 
 [tool.ruff]
-select = [
-    "E", "F", "W",  # default Flake8
+src = ["src"]
+
+[tool.ruff.lint]
+extend-select = [
     "I",            # isort
     "ISC",          # flake8-implicit-str-concat
     "PGH",          # pygrep-hooks
     "RUF100",       # unused noqa (yesqa)
     "UP",           # pyupgrade
+    "W",            # pycodestyle warnings
 ]
-src = ["src"]
 
 [tool.mypy]
 strict = true
 
 [tool.pytest.ini_options]
-addopts = "-rsx --tb=short"
+addopts = "-rsfE --tb=short"
 testpaths = ["tests"]
 
 [coverage.run]
@@ -91,13 +93,10 @@ relative_files = true
 show_missing = true
 
 [tool.tox]
-legacy_tox_ini = """
-[tox]
-envlist = py38, py39, py310, py311, py312
+env_list = ["py39", "py310", "py311", "py312", "py313"]
 skip_missing_interpreters = true
-minversion = 4.0.0
 
-[testenv]
-extras = test
-commands = python -m pytest {posargs}
-"""
+[tool.tox.env_run_base]
+package = "editable"
+commands = [["python", "-m", "pytest", { replace = "posargs", extend = true }]]
+extras = ["test"]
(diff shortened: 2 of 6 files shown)


agronholm:
This needs a regression test and a changelog entry.

sheinbergon:
> This needs a regression test and a changelog entry.

OK, I'll add the changes to the Changes.rst

As for a regression test - there's already a test that generates an enum (an intiial trial attempt caused it to fail).
However, like all of your tests, it doesn't really generates code a live db, but rather from an SQLalchemy definition set, 

so it was passing, while differing from real world behavior which caused a differen enum statement to be generated. How would you like to tackle that? 

I can see if maybe a specific postgresql.ENUM dialect would trigger the error

agronholm:
> > This needs a regression test and a changelog entry.
> 
> OK, I'll add the changes to the Changes.rst
> 
> As for a regression test - there's already a test that generates an enum (an intiial trial attempt caused it to fail). However, like all of your tests, it doesn't really generate code from a live db, but rather from an SQLalchemy definition set,
> 
> so it was passing all along, while differing from real world behavior which caused a differen enum statement to be generated. How would you like to tackle that?
> 
> I can see if maybe a specific postgresql.ENUM dialect would trigger the error

Yes, that would probably do the trick. IIRC I've used this technique elsewhere in the test suite when the generic types weren't reproducing the issues I wanted to fix.

sheinbergon:
> > > This needs a regression test and a changelog entry.
> > 
> > 
> > OK, I'll add the changes to the Changes.rst
> > As for a regression test - there's already a test that generates an enum (an intiial trial attempt caused it to fail). However, like all of your tests, it doesn't really generate code from a live db, but rather from an SQLalchemy definition set,
> > so it was passing all along, while differing from real world behavior which caused a differen enum statement to be generated. How would you like to tackle that?
> > I can see if maybe a specific postgresql.ENUM dialect would trigger the error
> 
> Yes, that would probably do the trick. IIRC I've used this technique elsewhere in the test suite when the generic types weren't reproducing the issues I wanted to fix.

@agronholm  so, I took a shot at it. 

This is not reproducible with the current set of unit tests (not for the `tables`, nor for the `declarative` generator). 
Even if I define PostgreSQL `ENUM` dialect type. When the `Table` is rendered and metadata is being loaded with the column definitions, it always `downgrades`  the dialect specific ENUM type to a plain SQLAlchemy `Enum` type, thus skipping this entire error. 

I can confirm my fix solves the issues for both `declarative` and `tables`  (as they both share the same function). 

The only way I see to do a regression test for something like is by doing real integration / e2e tests against a dockerized PostgreSQL DB. But that's going to be a whole lot of work for a one line fix ( perhaps worth considering for the future)

So, adding a test brings 0 value ATM on top of the current "fancy_types" one. If you know of way to have the dialect specific ENUM type survive metadata reflection, please let me know.

If not, I suggest we merge this fix without adding a test.

Added changes to change log, as you requested.

agronholm:
> > > > This needs a regression test and a changelog entry.
> > > 
> > > 
> > > OK, I'll add the changes to the Changes.rst
> > > As for a regression test - there's already a test that generates an enum (an intiial trial attempt caused it to fail). However, like all of your tests, it doesn't really generate code from a live db, but rather from an SQLalchemy definition set,
> > > so it was passing all along, while differing from real world behavior which caused a differen enum statement to be generated. How would you like to tackle that?
> > > I can see if maybe a specific postgresql.ENUM dialect would trigger the error
> > 
> > 
> > Yes, that would probably do the trick. IIRC I've used this technique elsewhere in the test suite when the generic types weren't reproducing the issues I wanted to fix.
> 
> @agronholm so, I took a shot at it.
> 
> This is not reproducible with the current set of unit tests (not for the `tables`, nor for the `declarative` generator). Even if I define PostgreSQL `ENUM` dialect type. When the `Table` is rendered and metadata is being loaded with the column definitions, it always `downgrades` the dialect specific ENUM type to a plain SQLAlchemy `Enum` type, thus skipping this entire error.
> 
> I can confirm my fix solves the issues for both `declarative` and `tables` (as they both share the same function).
> 
> The only way I see to do a regression test for something like is by doing real integration / e2e tests against a dockerized PostgreSQL DB. But that's going to be a whole lot of work for a one line fix ( perhaps worth considering for the future)
> 
> So, adding a test brings 0 value ATM on top of the current "fancy_types" one. If you know of way to have the dialect specific ENUM type survive metadata reflection, please let me know.
> 
> If not, I suggest we merge this fix without adding a test.
> 
> Added changes to change log, as you requested.

Thanks. As for the dialect specific enum type, aren't we skipping reflection and creating a metadata directly in the test? So basically we're creating a metadata object as if it had come back from SQLAlchemy reflection. What is happening there then that is different from what happens with a real PostgreSQL server?

sheinbergon:
> > > > > This needs a regression test and a changelog entry.
> > > > 
> > > > 
> > > > OK, I'll add the changes to the Changes.rst
> > > > As for a regression test - there's already a test that generates an enum (an intiial trial attempt caused it to fail). However, like all of your tests, it doesn't really generate code from a live db, but rather from an SQLalchemy definition set,
> > > > so it was passing all along, while differing from real world behavior which caused a differen enum statement to be generated. How would you like to tackle that?
> > > > I can see if maybe a specific postgresql.ENUM dialect would trigger the error
> > > 
> > > 
> > > Yes, that would probably do the trick. IIRC I've used this technique elsewhere in the test suite when the generic types weren't reproducing the issues I wanted to fix.
> > 
> > 
> > @agronholm so, I took a shot at it.
> > This is not reproducible with the current set of unit tests (not for the `tables`, nor for the `declarative` generator). Even if I define PostgreSQL `ENUM` dialect type. When the `Table` is rendered and metadata is being loaded with the column definitions, it always `downgrades` the dialect specific ENUM type to a plain SQLAlchemy `Enum` type, thus skipping this entire error.
> > I can confirm my fix solves the issues for both `declarative` and `tables` (as they both share the same function).
> > The only way I see to do a regression test for something like is by doing real integration / e2e tests against a dockerized PostgreSQL DB. But that's going to be a whole lot of work for a one line fix ( perhaps worth considering for the future)
> > So, adding a test brings 0 value ATM on top of the current "fancy_types" one. If you know of way to have the dialect specific ENUM type survive metadata reflection, please let me know.
> > If not, I suggest we merge this fix without adding a test.
> > Added changes to change log, as you requested.
> 
> Thanks. As for the dialect specific enum type, aren't we skipping reflection and creating a metadata directly in the test? So basically we're creating a metadata object as if it had come back from SQLAlchemy reflection. What is happening there then that is different from what happens with a real PostgreSQL server?

I have no idea about what might be different. But that's deep in SQLAlchemy territory. I think it's wrong to assume the test method starts from "here's a database reflected type", as it assumes to know a lot about SA internals, and **you're always testing without running a crucial part of the real-world flow which actually connects to a database**. 

Again - technically speaking - this was already covered in a test, except for the fact that test was always false positive

The correct approach would be to add real world integration tests, those will **NEVER** provide false positives as long as we test against a wide range of RDBMSs and versions. I suggest we open another issue for that and move forward.

sheinbergon:
@agronholm any response, I'd really like to close this issue and move to the next one. We have a milestone to reach!

agronholm:
I'd like to take a stab at researching why the test passes even with a native postgresql Enum type. I'll do this over the weekend, and if I'm none the wiser even after that, I'm okay with moving on.

sheinbergon:
Will be fixed in another PR by @agronholm

PR #366, opened 2025-02-13
adding arg to create model classes for tables w/o primary keys

pwyllcrusader:
In order to maintain some existing code there is a need to create Model Classes for all tables in database. So here I added a simple option to create Model Classes for every table

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -725,7 +725,7 @@ class DeclarativeGenerator(TablesGenerator):
         "use_inflect",
         "nojoined",
         "nobidi",
-        "model_classes"
+        "model_classes",
     }
 
     def __init__(
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -725,6 +725,7 @@ class DeclarativeGenerator(TablesGenerator):
         "use_inflect",
         "nojoined",
         "nobidi",
+        "model_classes"
     }
 
     def __init__(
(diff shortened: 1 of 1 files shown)


pwyllcrusader:
> But model classes will NOT work without primary keys, so what is this meant for?

well, I'm trying to replace some existing ORM classes with autogenerated ones. I didn't check every of them in use but I bet I saw such cases. What do you mean buy they didn't work?

agronholm:
See [here](https://docs.sqlalchemy.org/en/20/orm/quickstart.html#declare-models). In particular:
> All ORM mapped classes require at least one column be declared as part of the primary key, typically by using the [Column.primary_key](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.params.primary_key) parameter on those [mapped_column()](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.mapped_column) objects that should be part of the key.

The SQLAlchemy session contains a mapping of mapped class instances by their primary key values. If there is no primary key, this mapping cannot be done and the class is invalid.

pwyllcrusader:
> See [here](https://docs.sqlalchemy.org/en/20/orm/quickstart.html#declare-models). In particular:
> 
> > All ORM mapped classes require at least one column be declared as part of the primary key, typically by using the [Column.primary_key](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.params.primary_key) parameter on those [mapped_column()](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.mapped_column) objects that should be part of the key.
> 
> The SQLAlchemy session contains a mapping of mapped class instances by their primary key values. If there is no primary key, this mapping cannot be done and the class is invalid.

ok, sorry for doing things in a hurry and bothering you. I have to take a closer look on the task.

PR #385, opened 2025-03-10
🐛 Handles incomplete list-type

### LINKED ISSUE #383: Problem with Array generic type mapping
Hello everyone,

Just found this wonderful library, however I'm facing an issue that I couldn't find an answer for in the docs.

I have a table that uses postgresql ARRAYs, sqlacodegen is perfectly capable of creating the right side of the model, but the types are wrong (partially unknown to be precise). 

Anyone know a solution for this? I decided to open a discussion first instead of an issue since others might have already faces this.

```sql

CREATE TABLE FOO (
 id SERIAL PRIMARY KEY,
 some_ints INT[],
 some_floats FLOAT[]
);
```

```python
class Foo(Base)
 #...etc
 # right side is perfect! The type however is partially known
 some_ints: Mapped[Optional[list]] = mapped_column(ARRAY(Integer()))
```

Saw that this codebase is looking for maintainers, so if PRs are welcome I could take a crack at it.

_Originally posted by @JoaquimEsteves in https://github.com/agronholm/sqlacodegen/discussions/382_

JoaquimEsteves:
Closes #383

Note: I don't have a postgres DB to play with this patch - I'm kind of hopping that the tests are enough. I can report back tomorrow when I'm back at $WORK

Another thing I could do (since I have my hands in the dough) is to add an option to the cli that makes it so that we don't import `typing.List`, we could very well get away with just using `list[<attribute here>]`.

Indeed `typing.List` has been deprecated [since python 3.9](https://docs.python.org/3/library/typing.html#typing.List).
I could also do the same for `Optional` but maybe that's best left to some other PR.


src/sqlacodegen/generators.py
```diff
@@ -1219,7 +1219,7 @@ def recursively_get_col_type(column_type: TypeEngine[Any]) -> str:
                 python_type_module = python_type.__module__
                 self.add_module_import(python_type_module)
                 return f"{python_type_module}.{python_type_name}"
-            except NotImplementedError as wtf:
```
JoaquimEsteves: Nobody saw this :sweat_smile:

src/sqlacodegen/generators.py
```diff
@@ -1201,18 +1202,28 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
         column = column_attr.column
         rendered_column = self.render_column(column, column_attr.name != column.name)
 
-        try:
-            python_type = column.type.python_type
+        def recursively_get_col_type(column_type: TypeEngine[Any]) -> str:
```
JoaquimEsteves, line 1205: Thinking about it a little further doing a whole recursive clojure is a bit...too much.

It sure looks neat, but I can make it a more boring `if` condition if the maintainers think it's a good idea.

sheinbergon: Let's start with a boring `if` variation and move from there. SQL (PG) arrays is a very specific beast.  I don't see us currently supporting complex collections with nested generic types (not really a thing in RDBMSs)

src/sqlacodegen/generators.py
```diff
@@ -1201,18 +1202,28 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
...
+            if column_type.python_type == list:
+                self.add_literal_import("typing", "List")
+                # Dimensions is how postgres handles matrices
```
sheinbergon, line 1208: No need for these comments. We avoid relating to implementation specific details as much as possible in these areas of the code. Also, `dimensions` is supported by the base sa `ARRAY` type

src/sqlacodegen/generators.py
```diff
@@ -1201,18 +1202,28 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
...
+                # See: https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#sqlalchemy.dialects.postgresql.ARRAY
+                dim = getattr(column_type, "dimensions", None) or 1
+
+                return f"{'List[' * dim}{recursively_get_col_type(column_type.item_type)}{']' * dim}"
```
sheinbergon, line 1212: As you suggested, and as python 3.9 is now the minimal supported version, we can indeed get rid of `typing.List` and replace it with the standard `list`

sheinbergon (changes requested):
Thank you very much for your contribution. The result creates a conflicting import where sqlalchemy ARRAY and postgres dialect ARRAY, are imported. You need to adapt your code to make sure only one type surfaces

JoaquimEsteves:
Thank you for your review. I'll implement the necessary changes tomorrow evening.

JoaquimEsteves:
Should have addressed all of the points.

* :scissors:  `typing.List` is gone
* :scissors:  new comments are gone
*  Simplified the function; although looking at it I kind of prefer the old recursive function, because there's a `pre` and a `post-script`  component to the wrapped type I couldn't think of a nicer way to structure the function. Thought about making it an iterator or making the `column_python_type` a list that we can then join at the end, but wasn't happy with either result and so went for the most boring one. If you can think of anything better feel free to suggest or edit the code directly.

Note: That the bug [mentioned here](https://github.com/agronholm/sqlacodegen/pull/385#discussion_r1987934607) seems to have been part of the code for a while, the tests just never checked for an import with the same name from some other module - probably because it just doesn't happen.
I thought about fixing it changing the way the imports are handled (if `ARRAY` is already in the name-space give it another name, but then hmmm, what name could that be?) .
I think that that might be slightly out of scope for what I wanted.

src/sqlacodegen/generators.py
sheinbergon: 10x

src/sqlacodegen/generators.py
```diff
@@ -1201,18 +1202,28 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
         column = column_attr.column
         rendered_column = self.render_column(column, column_attr.name != column.name)
 
-        try:
-            python_type = column.type.python_type
+        def recursively_get_col_type(column_type: TypeEngine[Any]) -> str:
+            if column_type.python_type == list:
```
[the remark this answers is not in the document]
sheinbergon: great

src/sqlacodegen/generators.py
sheinbergon: great

tests/test_generator_declarative.py
```diff
@@ -1509,3 +1510,38 @@ class Simple(Base):
...
+        # TODO: How do the others handle multi-dimensional arrays?
+        Column("str_matrix", postgresql.ARRAY(VARCHAR(), dimensions=2)),
+    )
+
+    validate_code(
+        generator.generate(),
+        """\
+from typing import List, Optional
+
+from sqlalchemy import ARRAY, INTEGER, Integer, VARCHAR
```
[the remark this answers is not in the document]
sheinbergon: @JoaquimEsteves even if this was the case up till now, leaving something broken is not a good idea. It's not about handling imports differently, it's about type compilation. We should not use postgresql dialect arrays, as vannial sqlalchemy arrays provide the same functionality. If you need help finding where to look in the code to make sure arrays get the same treatment, let me know.

sheinbergon:
> Should have addressed all of the points.
> 
> * ✂️  `typing.List` is gone
> * ✂️  new comments are gone
> * Simplified the function; although looking at it I kind of prefer the old recursive function, because there's a `pre` and a `post-script`  component to the wrapped type I couldn't think of a nicer way to structure the function. Thought about making it an iterator or making the `column_python_type` a list that we can then join at the end, but wasn't happy with either result and so went for the most boring one. If you can think of anything better feel free to suggest or edit the code directly.
> 
> Note: That the bug [mentioned here](https://github.com/agronholm/sqlacodegen/pull/385#discussion_r1987934607) seems to have been part of the code for a while, the tests just never checked for an import with the same name from some other module - probably because it just doesn't happen. I thought about fixing it by changing the way the imports are handled (if `ARRAY` is already in the name-space give it another name, but then hmmm, what name could that be?) . I think that that might be slightly out of scope for what I wanted.


Again, thank you for your efforts. The PR is much better. But, we cannot merge it while the dual `ARRAY`  import remains.
If you need more help working on it, I can provide guidance. If you want me to "take over" and try and tackle it myself, let me know.

src/sqlacodegen/generators.py
sheinbergon: Great :clap:

src/sqlacodegen/generators.py
```diff
@@ -1201,18 +1202,29 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
         column = column_attr.column
         rendered_column = self.render_column(column, column_attr.name != column.name)
 
-        try:
-            python_type = column.type.python_type
+        def render_col_type(column_type: TypeEngine[Any]) -> str:
+            pre = ""
```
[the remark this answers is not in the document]
JoaquimEsteves: Sounds good to me. Just got a question: Would you rather this be a method of the generator or a pure-function on the utils? Because the current implementation doesn't reply on the `generator` class itself.

I'm thinking something like:

```python
@staticmethod # or it would just go on the `utils.py` file
def get_type_qualifiers(column_type: TypeEngine[Any]) -> tuple[str, TypeEngine[Any], str]:
  if isinstance(column_type, ARRAY):
     # ...etc
     return "list[", column_type.item_type, "]"
  return "", column_type, ""
```

JoaquimEsteves:
> But, we cannot merge it while the dual `ARRAY` import remains. If you need more help working on it, I can provide guidance. If you want me to "take over" and try and tackle it myself, let me know.

I can certainly take another crack at it.

Although I do think I would need some help on the dual `ARRAY` imports. When I tested it out at $WORK it didn't seem to have a double import, so I just thought it was bad test-data.

It could very well be a "skill-issue" on my side, I'm not quite familiar with the code yet.

sheinbergon:
> > But, we cannot merge it while the dual `ARRAY` import remains. If you need more help working on it, I can provide guidance. If you want me to "take over" and try and tackle it myself, let me know.
> 
> I can certainly take another crack at it.
> 
> Although I do think I would need some help on the dual `ARRAY` imports. When I tested it out at $WORK it didn't seem to have a double import, so I just thought it was bad test-data.
> 
> It could very well be a "skill-issue" on my side, I'm not quite familiar with the code yet.

Once you're done with the final fix suggested, let me know and I will take a look

JoaquimEsteves:
> Once you're done with the final fix suggested, let me know and I will take a look

Done!

src/sqlacodegen/generators.py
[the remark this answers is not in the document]
sheinbergon: The finalized version is bit too convoluted, but merhod separation make it more easily digestible. Fine by me

sheinbergon:
@

> > Once you're done with the final fix suggested, let me know and I will take a look
> 
> Done!
> 
> Edit: Feel free to ping me when you've gotten closer to the double-array problem. The closest I got was the comment on line `src/sqlacodegen/generators.py:272` telling me that the types have already been adapted - I just can't figure out where 😅
> 
> I think the following test demonstrates the problem.
> 
> ```python
> # Fails because postgres_array adds a second ARRAY import
> def test_table_with_arrays(generator: CodeGenerator) -> None:
>     _ = Table(
>         "with_items",
>         generator.metadata,
>         Column("id", INTEGER, primary_key=True),
>         Column("int_items_not_optional", ARRAY(INTEGER()), nullable=False),
>         Column("postgres_array", postgresql.ARRAY(INTEGER()), nullable=False),
>         Column("str_matrix", ARRAY(VARCHAR(), dimensions=2)),
>     )
> 
>     validate_code(
>         generator.generate(),
>         """\
> from typing import Optional
> 
> from sqlalchemy import ARRAY, INTEGER, Integer, VARCHAR
> from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column
> 
> class Base(DeclarativeBase):
>     pass
> 
> 
> class WithItems(Base):
>     __tablename__ = 'with_items'
> 
>     id: Mapped[int] = mapped_column(Integer, primary_key=True)
>     int_items_not_optional: Mapped[list[int]] = mapped_column(ARRAY(INTEGER()))
>     postgres_array: Mapped[list[int]] = mapped_column(ARRAY(INTEGER()))
>     str_matrix: Mapped[Optional[list[list[str]]]] = mapped_column(ARRAY(VARCHAR(), dimensions=2))
> """,
>     )
> ```

Maybe I'm not seeing things correctly, but after the recent iteration, I'm not seeing the double ARRAY import anymore. I'm also not seeing any misplaced fixes. @JoaquimEsteves am I missing anything?

tests/test_generator_declarative.py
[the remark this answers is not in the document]
sheinbergon: It looks good now

sheinbergon:
@agronholm this went some review iteration. I can pretty much approve what's going here. Please have another look

src/sqlacodegen/generators.py
```diff
@@ -1201,22 +1202,39 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
...
+                return python_type_name
+            try:
                 self.add_module_import(python_type_module)
-        except NotImplementedError:
-            self.add_literal_import("typing", "Any")
-            column_python_type = "Any"
+                return f"{python_type_module}.{python_type_name}"
+            except NotImplementedError:
+                self.add_literal_import("typing", "Any")
+                return "Any"
```
agronholm, line 1233: Why are we not reusing the `render_column_type()` method?

src/sqlacodegen/generators.py
```diff
@@ -1201,22 +1202,39 @@ def render_column_attribute(self, column_attr: ColumnAttribute) -> str:
...
-            python_type = column.type.python_type
+        def get_type_qualifiers() -> tuple[str, TypeEngine[Any], str]:
+            column_type = column.type
+            pre: list[str] = []
+            post_size = 0
+            if column.nullable:
+                self.add_literal_import("typing", "Optional")
+                pre.append("Optional[")
+                post_size += 1
+            if isinstance(column_type, ARRAY):
```
agronholm, line 1213: Add a blank line after a control block.
```suggestion

            if isinstance(column_type, ARRAY):
```

agronholm (changes requested):
What bugs me is the inconsistency. We're now suddenly using `list` but only in this new code. If we switch to using the generic built-ins, we should do the switch everywhere. And why did you duplicate the column type rendering code?

JoaquimEsteves:
Hello, thanks for the second look.

>  We're now suddenly using list but only in this new code. If we switch to using the generic built-ins, we should do the switch everywhere

I second the idea that adding more generic built-ins can be done in future PRs. For example, I didn't change the Optional because the `T | None` syntax is python 3.10+ only. I wanted to focus only on lists since they were a very common occurrence in the databases I use.

I didn't re-use the `render_column_type` because it returns sqlalchemy types instead of python types; so `Mapped[Integer]` instead of the correct `Mapped[int]`. The function looked quite tricky, so I just wrapped the code that was on `render_column_attributes` into some function (I'll readily admit this could very much be a skill issue on my side).

The double array problem went away once I fixed the test data.  Running my patch on a postgres DB with arrays didn't yield any double imports.

agronholm:
> I didn't re-use the `render_column_type` because it returns sqlalchemy types instead of python types; so `Mapped[Integer]` instead of the correct `Mapped[int]`.

Right, of course. But it seems to me like this should be an overrideable method in the generator then, `render_annotation()` or something. We can probably refactor it that way in a follow-up, but for the moment please at least rename the local function to avoid any misconceptions about its purpose.

tests/test_generator_declarative.py
```diff
@@ -1509,3 +1501,34 @@ class Simple(Base):
 server_default=text("'test'"))
 """,
     )
+
+
+def test_table_with_arrays(generator: CodeGenerator) -> None:
+    _ = Table(
```
agronholm, line 1507: Why assign the table if you're not using the variable?
```suggestion
    Table(
```

JoaquimEsteves: The underscore assignment was just habit - my LSP of choice is basedpyright which I've setup to complain if some function returns a value that I then don't check or use. Thanks for correcting it.

JoaquimEsteves:
Ahoy hoy,

Renamed the local function from `render_col_type` -> `render_python_type`.

If there's anything else feel free to ping :)

sheinbergon:
@agronholm @JoaquimEsteves let's also document this change in `CHANGES.rst`?

agronholm:
> @agronholm @JoaquimEsteves let's also document this change in `CHANGES.rst`?

Oh, right, that's still missing.

JoaquimEsteves:
How about:

```rst
**3.?.?**

- Array types are no longer partially unknown.


```

I don't know which version - should I put `**UNRELEASED**` as the following commit did?

e34718c67d66573ca2258ef361cd58562b7e1baa

agronholm:
> How about:
> 
> ```
> **3.?.?**
> 
> - Array types are no longer partially unknown.

Perhaps something like "Type annotations for `ARRAY` column attributes now include the Python type of the array elements"

agronholm:
Great, works for me.

JoaquimEsteves:
Grand - I'll wait for @sheinbergon to accept, feel free to merge at your convenience.

sheinbergon (approved):
Thank you

Issue #392, opened 2025-04-04
All generators generate the same code


treeline-jacob:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### Sqlacodegen version

3.0.0

### SQLAlchemy version

2.0.37

### RDBMS vendor

PostgreSQL

### What happened?

Each of the following commands generates the same model code.
```
uv run sqlacodegen --schemas <schema>  --generator=tables  <db_url>
uv run sqlacodegen --schemas <schema>  --generator=declarative  <db_url>
uv run sqlacodegen --schemas <schema>  --generator=sqlmodels  <db_url>
uv run sqlacodegen --schemas <schema>  --generator=dataclasses  <db_url>
```

### Database schema for reproducing the bug

The schema + tables were created with `dbt` if that is a helpful data point

```
CREATE TABLE dbt_dev_group_members (
    id character varying,
    group_id character varying,
    microsoft_graph_group_id character varying,
    user_id character varying,
    microsoft_graph_user_id character varying,
    user_given_name character varying,
    user_surname character varying,
    user_email character varying,
    user_display_name character varying,
    group_display_name character varying,
    group_description character varying
);


CREATE TABLE dbt_dev_groups (
    id character varying,
    microsoft_graph_group_id character varying,
    display_name character varying,
    description character varying
);


CREATE TABLE dbt_dev_users (
    id character varying,
    microsoft_graph_user_id character varying,
    display_name character varying,
    given_name character varying,
    surname character varying,
    email character varying
);
```

sheinbergon:
@treeline-jacob so, it's something with the provided schema, it's the fact that `dbt` doesn't  generate primary keys.
If you go ahead and configure primary keys on your generated tables in postgresql, `sqlacodegen` will work as expected.

You should use dbt contracts to define primary keys, as described here  - https://discourse.getdbt.com/t/can-dbt-indicate-primary-keys/11064/2

@agronholm while I have no opinion about tables with or with primary key, we should probaly not generate broken bindings in case PK is absent. WDYT?

sheinbergon:
@agronholm would still love to hear you out on this issue

agronholm:
WDYM by "broken bindings"?

agronholm:
Great question. Probably not, so long as we emit warnings about tables without PK that were skipped.

sheinbergon:
I think it'll make this an easyfix. I'll try and have a look soon

PR #394, opened 2025-04-12
disable SQLModel code generation for tables without primary key, and resolve `__tablename__` pylance error

pcen:
## Changes

- fixes reportAssignmentType pylance error for `__tablename__` fields in generated code by adding `str` type annotation
- fixes SQLModel code gen when some but not all tables have no primary key by setting `metadata_ref` in `SQLModelGenerator`

## Checklist

- [X] You've added tests (in `tests/`) which would fail without your patch
- [X] You've added a new changelog entry (in `CHANGES.rst`).

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -782,7 +782,10 @@ class DeclarativeGenerator(TablesGenerator):
             for fk in list(other_table.foreign_keys):
                 if fk.column.table.name == table.name:
                     for constraint in list(other_table.constraints):
-                        if isinstance(constraint, ForeignKeyConstraint) and fk in constraint.elements:
+                        if (
+                            isinstance(constraint, ForeignKeyConstraint)
+                            and fk in constraint.elements
+                        ):
                             other_table.constraints.discard(constraint)
                     fk.parent.foreign_keys.discard(fk)
                     fks_to_remove.append(fk)
@@ -813,8 +816,13 @@ class DeclarativeGenerator(TablesGenerator):
         # them normally
         links: defaultdict[str, list[Model]] = defaultdict(lambda: [])
         for table in self.metadata.sorted_tables:
-            if not table.primary_key and not self.generate_tables_for_relations_without_pk:
-                self.warnings.append(f"table {table.name!r} has no primary key and will not be included in the generated models")
+            if (
+                not table.primary_key
+                and not self.generate_tables_for_relations_without_pk
+            ):
+                self.warnings.append(
+                    f"table {table.name!r} has no primary key and will not be included in the generated models"
+                )
                 self.remove_table_from_metadata(table)
                 continue
 
--- a/tests/test_generator_sqlmodel.py
+++ b/tests/test_generator_sqlmodel.py
@@ -210,4 +210,7 @@ def test_generate_table_without_id(
         """,
     )
     assert len(generator.warnings) == 1
-    assert generator.warnings[0] == "table 'test_table' has no primary key and will not be included in the generated models"
+    assert (
+        generator.warnings[0]
+        == "table 'test_table' has no primary key and will not be included in the generated models"
+    )


agronholm:
This is wrong. A unique index is not the same as a primary key! And what's with that `__tablename__` annotation? How is it that no other type checker has a problem with it? @sheinbergon thoughts?

agronholm:
OK, according to [this comment](https://github.com/agronholm/apscheduler/blob/master/src/apscheduler/datastores/sqlalchemy.py#L337), Pyright indeed seems to require a type annotation there. But ORM classes **require** primary keys, and generating code that indicates that a table has a primary key while it doesn't seems just plain wrong to me. This was discussed earlier [here](https://github.com/agronholm/sqlacodegen/issues/392#issuecomment-2790176888). Thus, given that SQLModel classes cannot be generated for such tables, my preferred solution is to skip them and emit warnings instead.

agronholm:
Correct me if I'm wrong, but SQLModel doesn't seem to play nice with plain Tables. I do agree that what it's doing right now is wrong, but generating model classes with wrong information isn't any better in my books.

agronholm:
Apologies, I misread the test code at first. The real question, then, is what users of SQLModel will expect. I'll wait for @sheinbergon to reply before deciding what to do. If the majority favors generating plain Tables instead of simply emitting warnings, then this PR looks good.

sheinbergon:
@pcen @agronholm I think users expecting  an `SQLModel` and getting a `Table` in return is somewhat of a WTF experience (certainly better than what the library does today, but still confusing, to say the least). 

If a user aims to generate an SQLModel, he should be able to succeed, or fail loudly trying to do so.

I'm still crossed about the scenarios of mixed tables (all of the tables are fine, but one doesn't have a PK).

Should he:

1. Fail completely?
2. Fallback to table generation for that specific table, but emit a warnning?
3. Skip the table and emit a warning?
4. Allow the user to sepcify his desired behavior?

Gut feeling is to go with option No.3. The ORM style for SQLModel doesn't really co-exists in the same spaces as table access, and if the user wants, he could generate tables separately from sqlmodels.

agronholm:
> @pcen @agronholm I think users expecting an `SQLModel` and getting a `Table` in return is somewhat of a WTF experience (certainly better than what the library does today, but still confusing, to say the least).
> 
> If a user aims to generate an SQLModel, he should be able to succeed, or fail loudly trying to do so.
> 
> I'm still crossed about the scenarios of mixed tables (all of the tables are fine, but one doesn't have a PK).
> 
> Should he:
> 
> 1. Fail completely?
> 2. Fallback to table generation for that specific table, but emit a warnning?
> 3. Skip the table and emit a warning?
> 4. Allow the user to sepcify his desired behavior?
> 
> Gut feeling is to go with option No.3. The ORM style for SQLModel doesn't really co-exists in the same spaces as table access, and if the user wants, he could generate tables separately from sqlmodels.

I think we're in agreement here. @pcen I think this would need a separate PR. I'm willing to accept the type annotation addition for `__tablename__` in this PR if you remove the unrelated changes.

sheinbergon:
> > @pcen @agronholm I think users expecting an `SQLModel` and getting a `Table` in return is somewhat of a WTF experience (certainly better than what the library does today, but still confusing, to say the least).
> > If a user aims to generate an SQLModel, he should be able to succeed, or fail loudly trying to do so.
> > I'm still crossed about the scenarios of mixed tables (all of the tables are fine, but one doesn't have a PK).
> > Should he:
> > 
> > 1. Fail completely?
> > 2. Fallback to table generation for that specific table, but emit a warnning?
> > 3. Skip the table and emit a warning?
> > 4. Allow the user to sepcify his desired behavior?
> > 
> > Gut feeling is to go with option No.3. The ORM style for SQLModel doesn't really co-exists in the same spaces as table access, and if the user wants, he could generate tables separately from sqlmodels.
> 
> I think we're in agreement here. @pcen I think this would need a separate PR. I'm willing to accept the type annotation addition for `__tablename__` in this PR if you remove the unrelated changes.

I 2nd that. @pcen Aside from that, notice that the `pre-commit` status check is complaining about something

agronholm:
> Makes sense, but regardless of what the plan is moving forward for sqlmodel generator returning `Table`, I think this PR should still be reviewed as-is. The change pertinent to fixing `Table` gen: https://github.com/agronholm/sqlacodegen/pull/394/files#diff-c3df7ac7c19cdbd4da1fd2beda91fb9bcb3a5ed35235f5a738a273342552405fL1403 is just setting `metadata_ref` which is currently the empty string?

I thought we made it clear that Tables should not be generated at all, but a warning emitted instead?

agronholm:
What I'm trying to tell you that instead of fixing the generated Tables for the SQLModel generator, we should change the behavior to instead drop those Tables on the floor and not generated any code for them, and instead emit a warning about the primary key missing.

agronholm:
I don't think it will matter either way if the `metadata_ref` is correct or not if we won't be generating tables from this generator anyway going forward. What does bother me a bit is the second changelog entry which might imply to some that the generator now works properly with tables w/o a PK (which it never will with SQLModel).

agronholm:
> if we skip the table without pk and emit a warning, we won't be able to generate models for other tables with relations to the table without pk.

Why? If we nix those tables from the `metadata`, sure we won't generate relationships to them, but otherwise it will work fine.

sheinbergon:
> I'm unable to re-produce the pre-commit mypy error locally. I'm running the same my version (1.15). Is there any steps documented to setting up the correct dev environment locally?

I'll check in the following day or so and let you know

agronholm:
> > I'm unable to re-produce the pre-commit mypy error locally. I'm running the same my version (1.15). Is there any steps documented to setting up the correct dev environment locally?
> 
> I'll check in the following day or so and let you know

Just run `pre-commit run -a` locally.

agronholm:
If that doesn't help, then perhaps this is somehow an issue of the Python version?

agronholm:
> @pcen so, looks like you went ahead and put revised feature, into this PR, even though @agronholm asked for the second feature to be covered in a separate PR. pre-commit is indeed failing on one of your additions due to problematic mypy typing. I would much rather see this as two separate PRs. @agronholm WDYT?

Indeed, two PRs would be better as they address different issues.

PR #411, opened 2025-07-09
Fixed same-name imports from wrong package

EthanKim8683:
## Changes

Fixes `add_import()` logic that incorrectly imports something from `sqlalchemy` if `sqlalchemy` has it. So, `UUID`, for example, is supposed to be imported from `uuid`, but `sqlalchemy` exports `UUID` as well, resulting in the original code importing from `sqlalchemy` instead.

Closes #415

## Checklist

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,8 @@ Version history
 - Prevent double pluralization (PR by @dkratzert)
 - Fixes DOMAIN extending JSON/JSONB data types (PR by @sheinbergon)
 - Temporarily restrict SQLAlchemy version to 2.0.41 (PR by @sheinbergon)
-- Fixes ``add_import`` behavior when adding imports from sqlalchemy and overall better alignment of import behavior(s) across generators
+- Fixes ``add_import`` behavior when adding imports from sqlalchemy and overall better
+  alignment of import behavior(s) across generators
 
 **3.0.0**
 


agronholm:
I at least would like to understand why this happens before we merge and move on.

agronholm:
First of all, the import grouping is wrong. It should group `uuid` higher as it's from the standard library. Second, why is it not doing `from uuid import UUID`?

EthanKim8683:
Ah, the reason why the declarative generator's output doesn't have this bug is because of its `render_python_type()` method, which adds the _module_ import for the type, and then uses the type like `f"{module}.{name}"` which avoids the mix-up.

EthanKim8683:
@sheinbergon I agree. It fixes naming collisions between imports too, which I've seen become a problem if you import both `uuid.UUID` and `sqlalchemy.UUID` as `UUID`.

sheinbergon:
Great. Would you be able to fix your PR accommodate that change?

EthanKim8683:
Hey guys, sorry about the wait. I've made the changes we talked about, plus a couple more. All in all, here's the list of changes:
- `add_import` checks whether imports belong to the `sqlalchemy` package before attempting to import from `sqlalchemy` (from before)
- Added `SQLModelGenerator` test for `uuid.UUID` (from before)
- `render_column_attribute` now consistently imports entire Python packages instead of individual imports from packages (as discussed)
- `group_imports` now groups module imports as well (as discussed, before it was just packages)
- `builtin_module_names` has been renamed to `stdlib_module_names` (based on the way it was used in `group_imports`, I figured it'd be more fitting)
- `stdlib_module_names` contains`sys.stdlib_module_names` if running Python 3.10 or higher, or gets the stdlib module names from `stdlib-list` otherwise.
- Fixed one `DataclassGenerator` test which didn't group module names correctly
-  Auto-incrementing primary keys are no longer optional in `SQLModelGenerator`, consistent with the other generators
- Updated `SQLModelGenerator` tests to reflect change above

EthanKim8683:
Happy to help :)

sheinbergon:
@agronholm can you please share your thoughts. I'd like to include this PR in the upcoming 3.1 release

agronholm:
> @EthanKim8683 almost forgot. Can you please add a short description to `CHANGES.rst`?

Yeah, it looks like they checked the box for an added changelog entry, without actually having done so.

sheinbergon:
@EthanKim8683  can we please follow up on 2 additional minor fixes so we can have this merged? I'd like this to be included in the upcoming release

sheinbergon:
@EthanKim8683 a kind reminder

PR #427, opened 2025-08-25
RFC: Add optional `--ignored-tables` argument

plaes:
## Changes

* Add optional `--ignored-tables [filename]` argument to specify file containing list of tables to be ignored.

## Checklist

- [ ] You've added tests (in `tests/`) which would fail without your patch

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Version history
 ===============
 
+
+**UNRELEASED**
+- Add optional `--ignored-tables` argument (PR by @plaes)
+
 **3.1.0**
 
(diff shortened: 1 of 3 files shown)


agronholm:
What is the use case where you have so many tables to ignore that it'd be impractical to list them on the command line or to just list the tables you do want?

sheinbergon:
@agronholm I don't mind accepting such a feature

agronholm:
At the very least the options need to be consistent. If `--tables` accepts a comma-separated list of tables but `--ignore-tables` (which would intuitively feel like its negative counterpart) accepts a path instead, they're not consistent.

PR #428, opened 2025-08-28
Support external python-types

### LINKED ISSUE #423: Possible issue with 3.1.0 and sqlacodegen[geoalchemy2]
### Things to check first

### Sqlacodegen version

3.1.0

### SQLAlchemy version

2.0.41

### RDBMS vendor

PostgreSQL

### What happened?

I have (unknowingly) upgraded to the latest version of sqlacodegen after a pip re-install and suddenly my export script was not working anymore. 
Doing a little bit of debugging it seems sqlalchemy was throwing from `type_api.py` line 643.
```python
    @property
    def python_type(self) -> Type[Any]:
        """Return the Python type object expected to be returned
        by instances of this type, if known.

        Basically, for those types which enforce a return type,
        or are known across the board to do such for all common
        DBAPIs (like ``int`` for example), will return that type.

        If a return type is not defined, raises
        ``NotImplementedError``.

        Note that any type also accommodates NULL in SQL which
        means you can also get back ``None`` from any type
        in practice.

        """
        raise NotImplementedError()
```
After checking further, it seems it is a `POINT` type that is throwing off sqlacodegen.

After rollback to 3.0.0, everything goes back to normal.

Let me know if you need more info.

```sql
CREATE TABLE public.geo_table (
	id serial4 NOT NULL,
	point public.geometry(point, 4326) NULL,
	polygon public.geometry(polygon, 4326) NULL,
	CONSTRAINT geo_table_pkey PRIMARY KEY (_id)
);
```

sheinbergon:
## Changes

Fixes #423

Adds a fallback for `NotImplemented` python_type errors, including test coverage using geoalchemy2

## Checklist

these instructions.

### Updating the changelog

```
```


sheinbergon:
@agronholm I'm working to fix python 3.9 compatbility. As this regression is breaking, I'd like to release it as 3.1.1 as soon as we merge. That's why the CHANGES.rst already maps to 3.1.1

sheinbergon:
@agronholm - The test does not cover python 3.9, as geoalchemy2 dropped support for python 3.9 in its latest release and changed some other default behavior. As python 3.9 will soon reach EOL, I so no point in trying to cover it.

src/sqlacodegen/generators.py
```diff
@@ -13,7 +13,7 @@
 from keyword import iskeyword
 from pprint import pformat
 from textwrap import indent
-from typing import Any, ClassVar, Literal, cast
+from typing import Any, ClassVar, Literal, _SpecialForm, cast
```
agronholm, line 16: I'm not super fond of importing undocumented classes. For the special case of `Any`, can we just hard-code its name?

CHANGES.rst
```diff
@@ -1,6 +1,11 @@
 Version history
 ===============
 
+**3.1.1**
+
+- Fallack ``NotImplemented`` errors encountered when accessing ``python_type`` for non-native types to ``typing.Any``
+  (PR by @sheinbergon, based on work by @danplischke)
```
agronholm, line 7: Fixed typo and text wrapping.
```suggestion
- Fallback ``NotImplemented`` errors encountered when accessing ``python_type`` for
  non-native types to ``typing.Any``
  (PR by @sheinbergon, based on work by @danplischke)
```

src/sqlacodegen/generators.py
```diff
@@ -1243,9 +1243,23 @@ def render_python_type(column_type: TypeEngine[Any]) -> str:
...
+                    return python_type.__name__
+                elif isinstance(python_type, _SpecialForm):
+                    return python_type._name
+                else:
+                    raise AttributeError(
+                        f"Cannot get name from type {python_type}: no __name__ attribute and not a _SpecialForm"
+                    )
 
-            python_type_name = python_type.__name__
+            python_type_name = get_python_type_name()
```
agronholm, line 1262: Why are we wrapping this logic in a nested function that only gets called once?

tests/test_generator_declarative.py
```diff
@@ -1706,3 +1709,46 @@ class TestDomainJson(Base):
...
+        Column("id", INTEGER, primary_key=True),
+        Column("geom", Geometry("POINT", srid=4326, dimension=2), nullable=False),
+        Column("geog", Geography("POLYGON", dimension=2)),
+    )
+
+    validate_code(
+        generator.generate(),
+        """\
+from typing import Any, Optional
+import typing
```
agronholm, line 1733: This makes zero sense. Why are we using both `Any` and `typing.Any`? We should only be using `Any`.

agronholm (changes requested):

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Version history
 ===============
 
+**3.1.1**
```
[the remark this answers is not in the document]
sheinbergon: We can set it later, but I see no point in doing so

src/sqlacodegen/generators.py
sheinbergon: Just for the sake of clarity. I can move it out of the function if you feel it is better to do so

tests/test_generator_declarative.py
sheinbergon: OK, I will consolidate this behavior with the rest of the code

src/sqlacodegen/generators.py
agronholm: See my suggestion below.

src/sqlacodegen/generators.py
sheinbergon: @agronholm I added your final suggestion. Giving it a second look, I was able to remove some additional duplicate handling

tests/test_generator_declarative.py
sheinbergon: Consolidated this behavior. only `typing.Any` is imported, which matches the behavior for other non primite types (like the library treats uuids, as an exmaple)

src/sqlacodegen/generators.py
sheinbergon: Suggestion accepeted. Resolving

tests/test_generator_declarative.py
```diff
@@ -1706,3 +1709,45 @@ class TestDomainJson(Base):
...
+
+class SpatialTable(Base):
+    __tablename__ = 'spatial_table'
+    __table_args__ = (
+        Index('idx_spatial_table_geog', 'geog'),
+        Index('idx_spatial_table_geom', 'geom')
+    )
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    geom: Mapped[typing.Any] = mapped_column(Geometry('POINT', 4326, 2, from_text='ST_GeomFromEWKT', name='geometry', nullable=False), nullable=False)
```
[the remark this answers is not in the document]
sheinbergon: Or this, in case we want to protect against weird built in accessor methods

```python
        def render_python_type(column_type: TypeEngine[Any]) -> str:
            if isinstance(column_type, DOMAIN):
                column_type = column_type.data_type

            try:
                python_type = column_type.python_type
                python_type_module = python_type.__module__
                python_type_name = python_type.__name__
            except NotImplementedError:
                self.add_literal_import("typing", "Any")
                return "Any"

            if python_type_module == "builtins":
                return python_type_name

            self.add_module_import(python_type_module)
            return f"{python_type_module}.{python_type_name}"


```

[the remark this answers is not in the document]
sheinbergon: Done

agronholm (approved):

sheinbergon:
@agronholm can you push the 3.1.1 tag?

sheinbergon:
I don't think I can. I'll try in the next release :pray:

agronholm:
You can try pushing a non-numeric tag, as those will not trigger the release process.

PR #440, opened 2025-12-01
WIP: Fix  ignoring kwargs and ensure deterministic output (#439)

### LINKED ISSUE #439: `render_index` method ignores reflected `kwargs`
### Things to check first

### Sqlacodegen version

sqlacodegen==3.2.0 (from git+https://github.com/agronholm/sqlacodegen.git@615d79c6f6fdcb776c891179eb46516b1e8fd6f5)

### SQLAlchemy version

sqlalchemy>=2.0.44

### RDBMS vendor

MySQL (or compatible)

### What happened?

While generating SQLAlchemy models, I found that the `render_index` method ignores the `kwargs` of reflected indexes, which results in missing index attributes in the generated code. The current implementation only hard-codes the `unique` argument, and I’m not sure whether this was an intentional design choice for specific scenarios. I encountered the following two issues:

* When the database specifies an index length, the generated model omits this length. If the column’s length exceeds the database’s default index length, the generated DDL fails to create the table.
* Full-text index parameters are also stored in `kwargs`, but the generated model ignores them. As a result, the generated DDL creates a regular index instead of a full-text index.

```python
from sqlacodegen.generators import TablesGenerator
from sqlalchemy import (
    Column,
    Index,
    Integer,
    MetaData,
    String,
    Table,
    create_engine,
    create_mock_engine,
)


# Create a MySQL engine (no actual connection needed for DDL generation)
def dump(sql, *multiparams, **params):
    print(sql.compile(dialect=engine.dialect))


# Real engine
engine = create_engine("mysql+pymysql://root:123456@127.0.0.1:3306/test")
mock_engine = create_mock_engine(
    "mysql+pymysql://root:123456@127.0.0.1:3306/test", executor=dump
)
metadata = MetaData()


# Define a table and index
t = Table(
    "t",
    metadata,
    Column("id", Integer, primary_key=True),
    Column(
        "name",
        String(1000),
    ),
)

# Index with dialect options
idx = Index("idx_col1", t.c.name, mysql_prefix="FULLTEXT", mysql_with_parser="ngram")
idx2 = Index("idx_col2", t.c.name, mysql_length={"name": 100})

metadata.create_all(mock_engine)  # show ddl

# Drop table if exists and create new one
metadata.drop_all(engine)
metadata.create_all(engine)


# Reflect the table back
metadata_reflected = MetaData()
metadata_reflected.reflect(bind=engine)

# Generator
generator = TablesGenerator(metadata_reflected, engine, [])

# Find the reflected table
t_reflected = metadata_reflected.tables["t"]

# Render
print("--- Reflected Indexes ---")
for index in t_reflected.indexes:
    print(f"Index name: {index.name}")
    print(f"Rendered: {generator.render_index(index)}")
    print(f"Kwargs: {dict(index.kwargs)}")
    print("-" * 20)
# output
# CREATE TABLE t (
#         id INTEGER NOT NULL AUTO_INCREMENT,
#         name VARCHAR(1000),
#         PRIMARY KEY (id)
# )


# CREATE INDEX idx_col2 ON t (name(100))
# CREATE FULLTEXT INDEX idx_col1 ON t (name) WITH PARSER ngram
# --- Reflected Indexes ---
# Index name: idx_col1
# Rendered: Index('idx_col1', 'name')
# Kwargs: {'mysql_prefix': 'FULLTEXT', 'mysql_with_parser': 'ngram'}
# --------------------
# Index name: idx_col2
# Rendered: Index('idx_col2', 'name')
# Kwargs: {'mysql_length': {'name': 100}}
# --------------------
```

Additionally, while investigating this issue, I noticed that another built-in method, `render_callable`, has a similar problem: when `kwargs` contain string values, the generated code does not include the necessary quotation marks, which causes syntax errors. The relevant code snippet is as follows (omitted):

```python

output = render_callable(
    "func1",
    "arg1",
    "arg2",
    kwargs={
        "kwarg1": 1,
        # args += tuple(f"{key}={value}" for key, value in kwargs.items()) 
        "kwarg2": "value2",
        "kwarg3": repr("value3"),
    },
    indentation="    ",
)

print(output)
```

CREATE TABLE t (
        id INTEGER NOT NULL AUTO_INCREMENT, 
        name VARCHAR(1000), 
        PRIMARY KEY (id)
)

CREATE INDEX idx_col2 ON t (name(100))
CREATE FULLTEXT INDEX idx_col1 ON t (name) WITH PARSER ngram

luliangce:
## Changes

Fixes #439.

This PR fixes an issue where `Index` kwargs (such as `postgresql_using`, `mysql_length`, etc.) were being ignored during code generation.

Key changes:
- Modified `render_index` in `src/sqlacodegen/generators.py` to include `kwargs` in the generated `Index` definition.
- Enforced sorting of `kwargs` keys using `sorted()` to ensure deterministic output (avoiding random reordering of arguments in generated code).
- Added a new test case `test_index_with_kwargs` in `tests/test_generator_declarative.py`.
- Updated existing GeoAlchemy2 tests in `tests/test_generator_declarative.py`. These tests previously passed because the generator ignored the `postgresql_using='gist'` argument present in the metadata.
```python
# https://github.com/geoalchemy/geoalchemy2/blob/0c311122e5228a885e262d8f79b037cb74c0949b/geoalchemy2/admin/dialects/postgresql.py#L38C1-L57C15
def create_spatial_index(bind, table, col):
    """Create spatial index on the given column."""
    if col.type.use_N_D_index:
        postgresql_ops = {col.name: "gist_geometry_ops_nd"}
    else:
        postgresql_ops = {}
    if _check_spatial_type(col.type, Raster):
        col_func = func.ST_ConvexHull(col)
    else:
        col_func = col
    idx = Index(
        _spatial_idx_name(table.name, col.name),
        col_func,
        postgresql_using="gist", 
        postgresql_ops=postgresql_ops,
        _column_flag=True,
    )
    if bind is not None:
        idx.create(bind=bind)
    return idx
```

## Checklist

- [ ] You've added a new changelog entry (in `CHANGES.rst`).

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -426,7 +426,7 @@ class TablesGenerator(CodeGenerator):
         extra_args = [repr(col.name) for col in index.columns]
         kwargs = {
             key: repr(value) if isinstance(value, str) else value
-            for key, value in sorted(index.kwargs.items())
+            for key, value in sorted(index.kwargs.items(), key=lambda item: item[0])
         }
         if index.unique:
             kwargs["unique"] = True


sheinbergon:
@agronholm wdyt?

agronholm (approved):
I'm fine with this, but I do wonder about the claim about randomness. Is this something that was actually observed to happen?

sheinbergon:
Hello again @luliangce 

Were you able to make any progress with your work?

sheinbergon:
@agronholm I wonder if we should take this PR as is. Though not complete, it adds a missing capability that would address most of the use cases. WDYT?

src/sqlacodegen/generators.py
```diff
@@ -424,7 +424,10 @@ def render_table(self, table: Table) -> str:
 
     def render_index(self, index: Index) -> str:
         extra_args = [repr(col.name) for col in index.columns]
-        kwargs = {}
+        kwargs = {
+            key: repr(value) if isinstance(value, str) else value
+            for key, value in sorted(index.kwargs.items())
```
agronholm:
```suggestion
            for key, value in sorted(index.kwargs.items(), key=lambda item: item[0])
```

PR #455, opened 2026-02-06
Support enum arrays

### LINKED ISSUE #454: Array of enums not generating Enums
### Things to check first

### Sqlacodegen version

4.0.0rc2

### SQLAlchemy version

v2.0.43

### RDBMS vendor

PostgreSQL

### What happened?

When generating models from types with enums as arrays, the type is not converted into an Enum:

```python
class MyTable(Base):
    __tablename__ = 'my_table'
    __table_args__ = (
        PrimaryKeyConstraint('id', name='my_table_pkey'),
        {'schema': 'c'}
    )

    id: Mapped[int] = mapped_column(Integer, primary_key=True)
    my_enum: Mapped[list[str]] = mapped_column(ARRAY(Enum('a', 'b', name='my_enum', schema='c', _create_events=False)), nullable=False)
```

I would expect it to generate something like:
```python
    my_enum: Mapped[list[MyEnum]] = mapped_column(ARRAY(Enum(MyEnum, schema='c', _create_events=False)), nullable=False)

```

```sql
CREATE SCHEMA c;
CREATE TYPE c.my_enum AS ENUM (
    'a',
    'b'
);
CREATE TABLE c.my_table (
    id INT PRIMARY KEY,
    my_enum c.my_enum[] NOT NULL
);
```

sheinbergon:
## Changes

Fixes #454

Add support for Enum arrays

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,23 +5,15 @@ Version history
 
 - **BACKWARD INCOMPATIBLE** Relationship names changed when multiple FKs or junction tables
   connect to the same target table. Regenerating models will break existing code.
-- Improved relationship naming: one-to-many uses FK column names (e.g.,
-  ``simple_items_parent_container``), many-to-many uses junction table names (e.g.,
-  ``students_enrollments``). Use ``--options nofknames`` to revert to old behavior. (PR by @sheinbergon)
-- Fixed ``Index`` kwargs (e.g. ``mysql_length``) being ignored during code generation
-  (PR by @luliangce)
-- Added support for generating Python enum classes for ``ARRAY(Enum(...))`` columns
-  (e.g., PostgreSQL ``ARRAY(ENUM)``). Supports named/unnamed enums, shared enums across
-  columns, and multi-dimensional arrays. Respects ``--options nonativeenums``.
 - Added support for generating Python enum classes for ``ARRAY(Enum(...))`` columns
   (e.g., PostgreSQL ``ARRAY(ENUM)``). Supports named/unnamed enums, shared enums across
   columns, and multi-dimensional arrays. Respects ``--options nonativeenums``.
   (PR by @sheinbergon)
 - Improved relationship naming: one-to-many uses FK column names (e.g.,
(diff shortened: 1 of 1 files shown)


src/sqlacodegen/generators.py
```diff
@@ -554,6 +554,19 @@ def render_column_type(self, column: Column[Any]) -> str:
...
+        if isinstance(column_type, ARRAY) and isinstance(column_type.item_type, Enum):
+            if enum_class_name := self.enum_classes.get(
+                (column.table.name, column.name)
+            ):
+                self.add_import(ARRAY)
+                self.add_import(Enum)
+                rendered_enum = f"Enum({enum_class_name}, values_callable=lambda cls: [member.value for member in cls])"
+                if column_type.dimensions is not None:
+                    kwargs["dimensions"] = repr(column_type.dimensions)
+                return render_callable("ARRAY", rendered_enum, kwargs=kwargs)
```
agronholm: Add a blank line after a control block.
```suggestion

                return render_callable("ARRAY", rendered_enum, kwargs=kwargs)
```

src/sqlacodegen/generators.py
```diff
@@ -884,6 +897,43 @@ def fix_column_types(self, table: Table) -> None:
                         )
                         self.enum_classes[(table.name, column.name)] = enum_class_name
 
+            # Handle ARRAY columns with Enum item types (e.g., PostgreSQL ARRAY(ENUM))
```
agronholm: Remove the blank line before this one, as the `elif` continues an existing control block.

tests/test_generator_tables.py
```diff
@@ -320,6 +320,89 @@ class StatusEnum(str, enum.Enum):
     )
 
 
+def test_array_enum_named(generator: CodeGenerator) -> None:
+    from sqlalchemy import Enum as SAEnum
+    from sqlalchemy.types import ARRAY
```
agronholm: Do these have to be inline imports?

agronholm (changes requested):
Mostly OK, save for the code duplication.

PR #468, opened 2026-02-28
Filter out empty dialect kwargs

### LINKED ISSUE #463: render_index emits empty dialect kwargs (e.g. postgresql_include=[])
### Things to check first

### Sqlacodegen version

sqlacodegen==4.0.0

### SQLAlchemy version

sqlalchemy>=2.0.44

### RDBMS vendor

PostgreSQL

### What happened?

After upgrading from 3.2.0 to 4.0.0 (to get the fix from #439/#440), I noticed that every index in the generated output now includes `postgresql_include=[]`, even when the INCLUDE list is empty (which is the default).

For example, a simple index generates as:

```python
Index("idx_cities_name_state", "name", "state", postgresql_include=[])
```

Instead of:

```python
Index("idx_cities_name_state", "name", "state")
```

This affects every index in the generated models, adding noise and creating unnecessarily large diffs.

### Root cause

In `render_index` ([generators.py L515-523](https://github.com/agronholm/sqlacodegen/blob/dbf0498/src/sqlacodegen/generators.py#L515-L523)), the kwargs dict comprehension passes through all `index.kwargs` without filtering empty containers:

```python
kwargs = {
    key: repr(value) if isinstance(value, str) else value
    for key, value in sorted(index.kwargs.items(), key=lambda item: item[0])
}
```

PostgreSQL's dialect reflection always includes `postgresql_include` in `index.kwargs`, even when the INCLUDE list is empty. Since there's no check for empty `list`/`dict` values, the empty list is rendered verbatim.

The same issue exists in `_add_dialect_kwargs_and_info` ([generators.py around L672-680](https://github.com/agronholm/sqlacodegen/commit/dbf0498#diff-c3df7ac7c19cdbd4da1fd2beda91fb9bcb3a5ed35235f5a738a273342552405fR672)), which handles dialect kwargs for tables and columns — it also doesn't filter out empty containers.

### Suggested fix

Filter empty `list` and `dict` values before adding them to kwargs:

```python
kwargs = {
    key: repr(value) if isinstance(value, str) else value
    for key, value in sorted(index.kwargs.items(), key=lambda item: item[0])
    if value not in ([], {})
}
```

And the same filter in `_add_dialect_kwargs_and_info`.

```sql
CREATE TABLE test_table (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL
);

CREATE INDEX idx_test_name ON test_table (name);
```

Then run:

```bash
sqlacodegen --generator declarative postgresql://user:pass@localhost/db
```

The generated `Index` will include `postgresql_include=[]` even though no INCLUDE clause was specified.

sheinbergon:
## Changes

Fixes #463 

Filter out empty dialect kwargs

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,13 +1,15 @@
 Version history
 ===============
 
-**Unreleased**
+**4.0.2**
 
 - Fixed rendering of inherited keyword arguments for dialect-specific types that use
   ``**kwargs`` in their initializers (such as MySQL ``CHAR`` with ``collation``) while
   preserving existing ``*args`` rendering behavior (PR by @hyoj0942)
 - Fixed missing metadata argument when rendering plain tables with the SQLModel
 - Added support for self-referential tables in the SQLModel generator (PR by @sheinbergon)
+- Fixed empty dialect kwargs (e.g. ``postgresql_include=[]``) being included in
+  rendered indexes, tables, and columns (PR by @sheinbergon)
 
 **4.0.1**
 
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -427,6 +427,7 @@ class TablesGenerator(CodeGenerator):
         kwargs = {
             key: repr(value) if isinstance(value, str) else value
             for key, value in sorted(index.kwargs.items(), key=lambda item: item[0])
+            if value not in ([], {})
         }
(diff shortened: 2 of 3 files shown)


sheinbergon:
@agronholm final fix to be included in `4.0,2`

PR #470, opened 2026-03-16
Improve Identity server_default rendering for Decimal values

NotCarlosSerrano:
## Changes

Improve handling of `Identity` server defaults when rendering columns.

Previously, `Identity` attributes containing `Decimal` values (e.g. `start`, `increment`, `minvalue`, `maxvalue`) were not rendered correctly. This change inspects the `Identity` parameters and explicitly renders non-default values. When a value is a `Decimal`, it is now emitted as `decimal.Decimal(...)` and the required module import is added.

This ensures generated models correctly represent database identity configurations that rely on `Decimal` values.

Fixes #.

## Checklist

- [X] You've added tests (in `tests/`) which would fail without your patch
- [X] You've added a new changelog entry (in `CHANGES.rst`).

these instructions.

### DIFF
--- a/tests/test_generator_tables.py
+++ b/tests/test_generator_tables.py
@@ -1223,8 +1223,8 @@ def test_identity_column_decimal_values(generator: CodeGenerator) -> None:
     # values instead of integers. This test ensures those are serialized correctly.
     identity = Identity(start=1, increment=2)
     # Simulate database reflection returning Decimal values (as MSSQL does)
-    setattr(identity, "start", Decimal("1"))
-    setattr(identity, "increment", Decimal("2"))
+    identity.start = Decimal("1")  # type: ignore[assignment]
+    identity.increment = Decimal("2")  # type: ignore[assignment]
     Table(
         "simple_items",
         generator.metadata,


tests/test_generator_tables.py
```diff
@@ -1217,6 +1218,38 @@ def test_identity_column(generator: CodeGenerator) -> None:
     )
 
 
+def test_identity_column_decimal_values(generator: CodeGenerator) -> None:
+    identity = Identity(start=1, increment=2)
+    # Simulate database reflection returning Decimal values
+    setattr(identity, "start", Decimal("1"))
+    setattr(identity, "increment", Decimal("2"))
```
agronholm: Why would you use `setattr()` here?
```suggestion
    identity.start = Decimal("1")
    identity.increment = Decimal("2")
```

src/sqlacodegen/generators.py
```diff
@@ -521,7 +522,29 @@ def render_column(
...
+            identity_kwargs: dict[str, Any] = {}
+
+            for name, param in inspect.signature(Identity).parameters.items():
+                if name == "self" or param.kind in (
+                    Parameter.VAR_POSITIONAL,
+                    Parameter.VAR_KEYWORD,
+                ):
+                    continue
+
+                value = getattr(identity, name, None)
```
sheinbergon: you can use walrus here

agronholm: I mulled that over and didn't bother mentioning it :) But either way works for me.

agronholm (approved):
LGTM. @sheinbergon feel free to merge unless you have more changes you'd like to see.

NotCarlosSerrano:
Thanks for the merge! 🙌
Do you know if there’s an upcoming release planned that will include this change? I’d like to start using it once it’s available.

PR #479, opened 2026-06-03
Add multi-file model output

### LINKED ISSUE #88: Output Models to Different Files
Feature Request: would be nice to have the option to output all of the model classes to separate files; saves having to do a lot of cut/pasting when the single output file is huge.

PaleNeutron:
Closes #88.

This adds a CLI option for writing generated models into a directory, with one Python file per generated model:

- `--output-directory <dir>` creates the directory if needed
- shared declarations are written to `base_model.py`
- an empty `__init__.py` is created
- each generated model is written to its own `.py` file

The generator API now also accepts `generate(multi_file=True)` and returns a mapping of module names to generated source strings. Calling `generate()` without arguments preserves the existing single-file behavior.

Verification:

- `uv run pytest`
- `uv run ruff check .`

### DIFF
--- a/src/sqlacodegen/cli.py
+++ b/src/sqlacodegen/cli.py
@@ -5,6 +5,7 @@ import ast
 import sys
 from contextlib import ExitStack
 from importlib.metadata import entry_points, version
+from pathlib import Path
 from typing import Any, TextIO
 
 from sqlalchemy.engine import create_engine
@@ -88,7 +89,14 @@ def main() -> None:
             "(values are parsed with ast.literal_eval)"
         ),
     )
-    parser.add_argument("--outfile", help="file to write output to (default: stdout)")
+    output_group = parser.add_mutually_exclusive_group()
+    output_group.add_argument(
+        "--outfile", help="file to write output to (default: stdout)"
+    )
+    output_group.add_argument(
+        "--output-directory",
+        help="directory to write generated models to, one file per model",
+    )
     args = parser.parse_args()
 
     if args.version:
@@ -132,6 +140,14 @@ def main() -> None:
         metadata.reflect(
             engine, schema, (generator.views_supported and not args.noviews), tables
         )
+
+    if args.output_directory:
+        output_directory = Path(args.output_directory)
+        output_directory.mkdir(parents=True, exist_ok=True)
+        for name, contents in generator.generate(multi_file=True).items():
+            (output_directory / f"{name}.py").write_text(contents, encoding="utf-8")
+
+        return
 
     # Open the target file (if given)
     with ExitStack() as stack:
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -14,7 +14,7 @@ from itertools import count
 from keyword import iskeyword
 from pprint import pformat
 from textwrap import indent
-from typing import Any, ClassVar, Literal, cast
+from typing import Any, ClassVar, Literal, cast, overload
 
 import inflect
 import sqlalchemy
@@ -110,8 +110,16 @@ class CodeGenerator(metaclass=ABCMeta):
     def views_supported(self) -> bool:
         pass
 
+    @overload
     @abstractmethod
-    def generate(self) -> str:
+    def generate(self, multi_file: Literal[False] = False) -> str: ...
+
+    @overload
+    @abstractmethod
+    def generate(self, multi_file: Literal[True]) -> dict[str, str]: ...
+
+    @abstractmethod
+    def generate(self, multi_file: bool = False) -> str | dict[str, str]:
         """
         Generate the code for the given metadata.
         .. note:: May modify the metadata.
@@ -167,10 +175,14 @@ class TablesGenerator(CodeGenerator):
             metadata_ref="metadata",
         )
 
-    def generate(self) -> str:
-        self.generate_base()
+    @overload
+    def generate(self, multi_file: Literal[False] = False) -> str: ...
 
-        sections: list[str] = []
+    @overload
+    def generate(self, multi_file: Literal[True]) -> dict[str, str]: ...
+
+    def generate(self, multi_file: bool = False) -> str | dict[str, str]:
+        self.generate_base()
 
         # Remove unwanted elements from the metadata
         for table in list(self.metadata.tables.values()):
@@ -199,6 +211,14 @@ class TablesGenerator(CodeGenerator):
         # Generate the models
         models: list[Model] = self.generate_models()
 
+        if multi_file:
+            return self.render_multi_file(models)
+        else:
+            return self.render_all(models)
+
+    def render_all(self, models: list[Model]) -> str:
+        sections: list[str] = []
+
         # Render module level variables
         if variables := self.render_module_variables(models):
             sections.append(variables + "\n")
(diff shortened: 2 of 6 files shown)


sheinbergon:
@PaleNeutron thank you for your contribution. While there has been a long discussion about this request in the original issue, I feel the a file per module maybe a bit of an exaggeration. I would say a file per schema makes more sense, but feel free to counter that idea. @agronholm wdyt?

PaleNeutron:
@sheinbergon , we subscribe financial data into a database and the provider put about 2000 tables in one schema. Split models by schema can not solve this problem.

BTW, generally schema count is less than 10 and generate each one in a loop is acceptable, no need to create a new option.

src/sqlacodegen/generators.py
```diff
@@ -220,6 +240,74 @@ def generate(self) -> str:
 
         return "\n\n".join(sections) + "\n"
 
+    def render_multi_file(self, models: list[Model]) -> dict[str, str]:
+        rendered_files = {
```
sheinbergon, line 244: I would rather much see the BaseModel inside `__init__.py`.  it's an sqlalchemy thing not a concrete database model

sheinbergon (changes requested):
One smal change request

sheinbergon:
> @sheinbergon , we subscribe financial data into a database and the provider put about 2000 tables in one schema. Split models by schema can not solve this problem.
> 
> BTW, generally schema count is less than 10 and generate each one in a loop is acceptable, no need to create a new option.

Alright, Fine by me. Submitted one small change request

src/sqlacodegen/generators.py
PaleNeutron: So in each model file, we would see something like `from .__init__ import Base`, I think it is a little strange.

Usually we collect submodules into `__init__.py` and export it to other module, but define base model in `__init__.py` is the reverse style.

But I agree with `base_model` should be different with database model. How about `__base_model.py`?

sheinbergon: Fair enough. Just `__base.py` or `__base__.py` would do

PaleNeutron: Updated to use `__base.py` and import from `.__base`.

src/sqlacodegen/generators.py
```diff
@@ -110,8 +110,16 @@ def __init__(
...
         pass
 
+    @overload
     @abstractmethod
-    def generate(self) -> str:
+    def generate(self, multi_file: Literal[False] = False) -> str: ...
+
+    @overload
+    @abstractmethod
+    def generate(self, multi_file: Literal[True]) -> dict[str, str]: ...
```
sheinbergon, line 119: Passing this flag a function argument is wrong. It also creates the need to overload  the `generate` method which serve as an entrypoint to the core functionality. Just include it as part of the constructor arg options and branch internally

sheinbergon (changes requested):
More changes requested

sheinbergon:
@agronholm aside from the additional comment - fine by me.  WDYT?

src/sqlacodegen/cli.py
```diff
@@ -133,6 +141,14 @@ def main() -> None:
             engine, schema, (generator.views_supported and not args.noviews), tables
         )
 
+    if args.output_directory:
```
sheinbergon, line 144: Better move this to live under the `ExitStack`

src/sqlacodegen/cli.py
```diff
@@ -133,6 +141,14 @@ def main() -> None:
...
+        output_directory = Path(args.output_directory)
+        output_directory.mkdir(parents=True, exist_ok=True)
+        for name, contents in generator.generate(multi_file=True).items():
```
sheinbergon, line 147: One of the things about code-generation libs was their ability to leave you in a broken state.
Sticking to a single module helped this library avoid that. With the introduction of multi-files, you need to start dealing with stuff like cleaning up residue from past versions and partial success, whih seems to be missing from the current implementaion

PaleNeutron: You are right, but what's the best implementation? Check the whole output dir and warning all redundant files?

PaleNeutron: We can delete the output dir and create a new one but I think it is a little crude. If user pass `/` to the output-directory arg...

sheinbergon: If the user is rackless, this piece of software is not going to be the one to save him

agronholm:
Multi-file generation is painfully hard to get right. Limiting it to file-per-schema doesn't really fix that, as foreign keys between schemas are a thing. What is your motivation here @PaleNeutron ?

agronholm:
> Multi-file generation is painfully hard to get right. Limiting it to file-per-schema doesn't really fix that, as foreign keys between schemas are a thing. What is your motivation here @PaleNeutron ?

Oh, I spotted your comment on the issue. Do you have thousands of tables in a single schema then?

PaleNeutron:
> > Multi-file generation is painfully hard to get right. Limiting it to file-per-schema doesn't really fix that, as foreign keys between schemas are a thing. What is your motivation here @PaleNeutron ?
> 
> Oh, I spotted your comment on the issue. Do you have thousands of tables in a single schema then?

Yes, a subscribed fin-data database.

agronholm:
I'm unfavorable to adding this to `sqlacodegen` due to the extra maintainer load it adds (not least due to the complexity of getting the table namings, relationships etc. right), to the benefit benefit of very few users (who always have an option of manually splitting the code).

PaleNeutron:
> I'm unfavorable to adding this to `sqlacodegen` due to the extra maintainer load it adds (not least due to the complexity of getting the table namings, relationships etc. right), to the benefit benefit of very few users (who always have an option of manually splitting the code).

I think this could still provide a noticeable benefit even for users with around 100 tables. Splitting the generated models into separate modules can improve startup performance quite a bit — in some cases saving around 1–5 seconds on every startup.

The fact that more users are not asking for this feature does not necessarily mean it would not be useful. A 1–5 second delay may be tolerable, but that does not mean the current approach is optimal.

I also do not think resolving relationships is especially complex in SQLAlchemy 2.0. We can use string-based references / forward annotations for relationships, for example:

```python
# Source - https://stackoverflow.com/a/79601366
# Posted by user15630736, modified by community. See post 'Timeline' for change history
# Retrieved 2026-06-07, License - CC BY-SA 4.0

import typing

if typing.TYPE_CHECKING:
  from users.models import User

class Account(BaseModel):
  __tablename__ = "accounts"

  user_id: Mapped[int] = mapped_column(ForeignKey("users.id"), nullable=False)
  user: Mapped["User"] = relationship(back_populates="accounts")

```

agronholm:
> > I'm unfavorable to adding this to `sqlacodegen` due to the extra maintainer load it adds (not least due to the complexity of getting the table namings, relationships etc. right), to the benefit benefit of very few users (who always have an option of manually splitting the code).
> 
> I think this could still provide a noticeable benefit even for users with around 100 tables. Splitting the generated models into separate modules can improve startup performance quite a bit — in some cases saving around 1–5 seconds on every startup.
> 
> The fact that more users are not asking for this feature does not necessarily mean it would not be useful. A 1–5 second delay may be tolerable, but that does not mean the current approach is optimal.

Sure, but I have to weigh the maintenance cost against the benefit, and I have no metric other than the number of users asking for this.

> 
> I also do not think resolving relationships is especially complex in SQLAlchemy 2.0. We can use string-based references / forward annotations for relationships, for example:

String-based references won't resolve unless the class in question has been imported. SQLAlchemy 2.0 did not change that fact. Lazy imports would handily fix this, but they were only recently introduced in Python 3.15 (still in beta).

agronholm:
> > I also do not think resolving relationships is especially complex in SQLAlchemy 2.0. We can use string-based references / forward annotations for relationships, for example:
> 
> String-based references won't resolve unless the class in question has been imported. SQLAlchemy 2.0 did not change that fact. Lazy imports would handily fix this, but they were only recently introduced in Python 3.15 (still in beta).

Actually, the mapper configuration step would result in all the lazy imports being triggered, making it pointless.

nerdstrike:
@agronholm parachuting into this PR/debate. Past experience with using DBIx::Class and its schema generator shows me that splitting models by file is helpful with complex schemata, but becomes more so if you checksum the resulting model and add a boundary beyond which customisations can be made. Then the generator (sqlacodegen or DBIx::Class::Schema::Loader) can safely update the model from a database while leaving supplemental code intact. The consumer of a schema can build up libraries of common code associated with each model without having to maintain multiple libraries/packages.

In my opinion this is the best way to do it, as it creates the smallest numbers of artifacts to maintain and keeps code close to the source - on the flip side it does increase the difficulty for... you.

"If you want custom code, you can just extend the models!" I hear you say. True, but super-not-fun if you've got hundreds of them and the original schema receives frequent changes.

I hope I've made you contemplate a change of heart, because we're getting plenty of value out of `sqlacodegen` even with the limitations.

sheinbergon:
> > > I'm unfavorable to adding this to `sqlacodegen` due to the extra maintainer load it adds (not least due to the complexity of getting the table namings, relationships etc. right), to the benefit benefit of very few users (who always have an option of manually splitting the code).
> > 
> > 
> > I think this could still provide a noticeable benefit even for users with around 100 tables. Splitting the generated models into separate modules can improve startup performance quite a bit — in some cases saving around 1–5 seconds on every startup.
> > The fact that more users are not asking for this feature does not necessarily mean it would not be useful. A 1–5 second delay may be tolerable, but that does not mean the current approach is optimal.
> 
> Sure, but I have to weigh the maintenance cost against the benefit, and I have no metric other than the number of users asking for this.

Looks like people are interested in this feature. Interested enough to comment/debate about it, that is
> 
> > I also do not think resolving relationships is especially complex in SQLAlchemy 2.0. We can use string-based references / forward annotations for relationships, for example:
> 
> String-based references won't resolve unless the class in question has been imported. SQLAlchemy 2.0 did not change that fact. Lazy imports would handily fix this, but they were only recently introduced in Python 3.15 (still in beta).

@agronholm I have to say - in the era of AI generated code, I don't find maintenance to be all that high. Maybe not in its current form, but I think we should find a way to enable this contribution

agronholm:
I'm willing to consider it if someone comes up with a working solution that doesn't horribly complicate our code.

PR #480, opened 2026-06-06
Preserve dialect-specific ARRAY types instead of adapting to generic ARRAY

### LINKED ISSUE #441: sqlacodegen generates base sqlalchemy.ARRAY for PostgreSQL text[], causing .contains() NotImplementedError (should use sqlalchemy.dialects.postgresql.ARRAY)
### Things to check first

### Sqlacodegen version

3.2.0

### SQLAlchemy version

2.0.44

### RDBMS vendor

PostgreSQL

### What happened?

When generating models from a PostgreSQL schema that contains an array column (e.g. test.tags text[]), sqlacodegen outputs from sqlalchemy import ARRAY and uses ARRAY(Text()) for the column type.
In SQLAlchemy 2.0, the base sqlalchemy.ARRAY does not implement PostgreSQL-specific array operators such as .contains(), so calling Model.tags.contains([...]) raises:

NotImplementedError: ARRAY.contains() not implemented for the base ARRAY type; please use the dialect-specific ARRAY type

This makes the generated model unusable for common PostgreSQL array queries unless the import/type is manually adjusted to sqlalchemy.dialects.postgresql.ARRAY.

1. Create a PostgreSQL table with an array column:

```
CREATE SCHEMA IF NOT EXISTS db;

CREATE TABLE db.test (
  id text PRIMARY KEY,
  tags text[]
);
```

2. Run sqlacodegen (declarative generator), e.g.:
`sqlacodegen postgresql://USER:PASSWORD@localhost:5432/DB --schemas db --generator declarative
`
3. Observe generated code includes base ARRAY import:

```
from sqlalchemy import ARRAY, Text
# ...
tags = mapped_column(ARRAY(Text()))

```
4. Run a query expression using .contains():

```
from database.models.models import test
Test.tags.contains(["hello"])
```

Raises:

NotImplementedError: ARRAY.contains() not implemented for the base ARRAY type; please use the dialect-specific ARRAY type

For PostgreSQL array columns (text[], int[], etc.), generated code should use the dialect-specific type:
from sqlalchemy.dialects.postgresql import ARRAY

# ...
tags = mapped_column(ARRAY(Text()))

so that .contains() (and other PostgreSQL array operators) work.

**Environment**

Python: 3.13
SQLAlchemy: >=2.0.44
sqlacodegen: >=3.2.0
Database: PostgreSQL
Column: Test.tags text[]

mokashang:
Fixes #441.

## Problem

When sqlacodegen reflects a PostgreSQL `text[]` (or any dialect-specific
ARRAY) column, `get_adapted_type` walks the column type's MRO and substitutes
the original `sqlalchemy.dialects.postgresql.ARRAY` with the generic
`sqlalchemy.ARRAY`. The generic ARRAY does not implement operators like
`.contains()`, `.any()` or `.all()`, so calling them on a generated model
raises at runtime:

```
NotImplementedError: ARRAY.contains() not implemented for the base ARRAY type;
please use the dialect-specific ARRAY type
```

The existing workaround (`--options keep_dialect_types`) opts out of *all*
generic-type adaptation, which is more than the user usually wants:
`Integer` becomes `INTEGER`, `Boolean` becomes `BOOLEAN`, and so on.

## Fix

`get_adapted_type` now special-cases ARRAY: when the input is a subclass of
the generic `ARRAY` (i.e. a dialect-specific ARRAY such as
`postgresql.ARRAY`), the original class is preserved and only the item type
is adapted recursively. Plain `sqlalchemy.ARRAY` inputs are unaffected, and
no other type's adaptation behavior changes.

This matches the direction the maintainer pointed to in the issue thread
("Should the array type perhaps be special cased? I think that overall the
generalization of types is correct.").

## Generated output before / after

Input column: `Column("tags", postgresql.ARRAY(postgresql.TEXT))`

Before:

```python
from sqlalchemy import ARRAY, Column, Table, Text

# tags.contains([...]) -> NotImplementedError at runtime
Column('tags', ARRAY(Text()))
```

After:

```python
from sqlalchemy import Column, Table, Text
from sqlalchemy.dialects.postgresql import ARRAY

# tags.contains([...]) -> 'tags @> :tags_1'
Column('tags', ARRAY(Text()))
```

Item types are still adapted: `postgresql.DOUBLE_PRECISION` still becomes
`Double`, `postgresql.TEXT` still becomes `Text`. Only the outer `ARRAY`
class is kept.

## Tests

- The existing `test_arrays` in `tests/test_generator_tables.py` is
  updated to expect `from sqlalchemy.dialects.postgresql import ARRAY`
  for inputs constructed with `postgresql.ARRAY(...)`. The item-type
  adaptation it was already exercising (`DOUBLE_PRECISION` -> `Double`,
  `INTEGER` -> `Integer`) is preserved.
- A new `test_array_preserves_dialect_for_runtime_operators` is added as
  a focused regression test that mirrors the scenario from the issue
  (`text[]` column) and documents *why* the dialect-specific ARRAY is
  required.
- Full test suite passes (`pytest -q` -> 157 passed). `ruff check`,
  `ruff format --check`, and `mypy src/sqlacodegen/generators.py` all
  pass.

## Notes / scope

- Tests that construct `ARRAY(SAEnum(...))` using the generic
  `sqlalchemy.ARRAY` directly continue to work unchanged, because the
  early-return only triggers when `type(coltype) is not ARRAY` (i.e.
  the input is a dialect-specific subclass).
- The `keep_dialect_types` option is unaffected; this change improves
  the default path, which is what the workaround was previously needed
  to escape.

### DIFF
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -1036,6 +1036,16 @@ class TablesGenerator(CodeGenerator):
                         column.server_default = None
 
     def get_adapted_type(self, coltype: Any) -> Any:
+        # The generic ``sqlalchemy.ARRAY`` does not implement operators such as
+        # ``.contains()`` that dialect-specific ARRAY types (notably
+        # ``sqlalchemy.dialects.postgresql.ARRAY``) provide. Adapting a
+        # dialect-specific ARRAY to the generic one would silently break user
+        # code that relies on those operators at runtime (see GH-441), so we
+        # keep the original ARRAY class and only adapt its item type.
+        if isinstance(coltype, ARRAY) and type(coltype) is not ARRAY:
+            coltype.item_type = self.get_adapted_type(coltype.item_type)
+            return coltype
+
         compiled_type = coltype.compile(self.bind.engine.dialect)
         for supercls in coltype.__class__.__mro__:
             if not supercls.__name__.startswith("_") and hasattr(
--- a/src/sqlacodegen/generators.py
+++ b/src/sqlacodegen/generators.py
@@ -1036,12 +1036,8 @@ class TablesGenerator(CodeGenerator):
                         column.server_default = None
 
     def get_adapted_type(self, coltype: Any) -> Any:
-        # The generic ``sqlalchemy.ARRAY`` does not implement operators such as
-        # ``.contains()`` that dialect-specific ARRAY types (notably
-        # ``sqlalchemy.dialects.postgresql.ARRAY``) provide. Adapting a
-        # dialect-specific ARRAY to the generic one would silently break user
-        # code that relies on those operators at runtime (see GH-441), so we
-        # keep the original ARRAY class and only adapt its item type.
+        # Keep dialect-specific ARRAY subclasses; the generic sqlalchemy.ARRAY
+        # is missing operators like .contains() (GH-441).
         if isinstance(coltype, ARRAY) and type(coltype) is not ARRAY:
             coltype.item_type = self.get_adapted_type(coltype.item_type)
             return coltype
--- a/tests/test_generator_tables.py
+++ b/tests/test_generator_tables.py
@@ -127,10 +127,15 @@ def test_arrays(generator: CodeGenerator) -> None:
         Column("int_array", postgresql.ARRAY(INTEGER)),
     )
 
+    # The dialect-specific ``postgresql.ARRAY`` is preserved because the generic
+    # ``sqlalchemy.ARRAY`` does not implement operators such as ``.contains()``
+    # (see GH-441). The item types are still adapted to their generic
+    # equivalents (``DOUBLE_PRECISION`` -> ``Double``, ``INTEGER`` -> ``Integer``).
     validate_code(
         generator.generate(),
         """\
-        from sqlalchemy import ARRAY, Column, Double, Integer, MetaData, Table
+        from sqlalchemy import Column, Double, Integer, MetaData, Table
+        from sqlalchemy.dialects.postgresql import ARRAY
 
         metadata = MetaData()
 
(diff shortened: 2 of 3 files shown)


tests/test_generator_tables.py
```diff
@@ -144,6 +149,43 @@ def test_arrays(generator: CodeGenerator) -> None:
     )
 
 
+@pytest.mark.parametrize("engine", ["postgresql"], indirect=["engine"])
+def test_array_preserves_dialect_for_runtime_operators(
+    generator: CodeGenerator,
+) -> None:
+    """Regression test for GH-441.
+
+    A ``text[]`` column should generate ``sqlalchemy.dialects.postgresql.ARRAY``
```
sheinbergon: We don't need these AI generated comments in test code. I hate Opus 4.8's wasteful blabbering

tests/test_generator_tables.py
```diff
@@ -127,10 +127,15 @@ def test_arrays(generator: CodeGenerator) -> None:
         Column("int_array", postgresql.ARRAY(INTEGER)),
     )
```
sheinbergon, line 129: Same

src/sqlacodegen/generators.py
```diff
@@ -1036,6 +1036,16 @@ def fix_enum_column(col_name: str, enum_type: Enum) -> None:
                         column.server_default = None
 
     def get_adapted_type(self, coltype: Any) -> Any:
+        # The generic ``sqlalchemy.ARRAY`` does not implement operators such as
```
sheinbergon: Make it more concise

sheinbergon:
@agronholm fine by me. WDYT?

src/sqlacodegen/generators.py
```diff
@@ -1036,6 +1036,16 @@ def fix_enum_column(col_name: str, enum_type: Enum) -> None:
...
+        # ``.contains()`` that dialect-specific ARRAY types (notably
+        # ``sqlalchemy.dialects.postgresql.ARRAY``) provide. Adapting a
+        # dialect-specific ARRAY to the generic one would silently break user
+        # code that relies on those operators at runtime (see GH-441), so we
+        # keep the original ARRAY class and only adapt its item type.
+        if isinstance(coltype, ARRAY) and type(coltype) is not ARRAY:
```
agronholm, line 1041: Is the `and type(coltype) is not ARRAY` part necessary? Will something break if it's removed?

agronholm:
> @agronholm fine by me. WDYT?

I share your sentiments about the excessively verbose AI comments. And I left a minor comment of my own.

src/sqlacodegen/generators.py
mokashang: Yes, it's necessary. The generic `sqlalchemy.ARRAY` still needs to fall through to the normal adaptation loop below — that loop's compile-equality check is what preserves the dialect-specific item type when `ARRAY(INTEGER)` is generated as `from sqlalchemy import ARRAY, INTEGER; ARRAY(INTEGER())` rather than `ARRAY(Integer())`.

I confirmed by removing the `and type(coltype) is not ARRAY` clause and running the suite: `test_table_with_arrays` and `test_array_enum_nonativeenums_option` both fail because the early return adapts the inner `INTEGER` to `Integer` (and `VARCHAR` to `String`), which the existing behavior on generic `ARRAY` declines to do.

Also addressed the comment-verbosity feedback in 32c2eed — trimmed the block comment here and the over-explanation in the test docstrings.

sheinbergon (approved):

PR #485, opened 2026-08-28
Fix remote_side for composite self-referential FK targeting a UNIQUE constraint (#484)

### LINKED ISSUE #484: Composite self-referential FK targeting a UNIQUE constraint generates wrong remote_side, mappers fail to configure
### Versions

- sqlacodegen 4.0.4
- SQLAlchemy 2.0.52
- PostgreSQL 17
- Python 3.13

### Summary

When a table has a **composite self-referential foreign key that targets a `UNIQUE` constraint rather than the primary key**, the generated `remote_side` names the primary key instead of the referenced columns. The module imports without error, and then mapper configuration fails on the first query, so an import-only smoke test does not catch it.

This looks adjacent to #233, but distinct: the fix there covers the self-reference that targets the PK, which still works correctly (contrast below).

### Reproducing

```sql
-- A: self-FK targeting the primary key. Generated code is correct.
CREATE TABLE node (
    id        bigserial PRIMARY KEY,
    parent_id bigint,
    CONSTRAINT node_parent_fkey FOREIGN KEY (parent_id) REFERENCES node (id)
);

-- B: composite self-FK targeting a UNIQUE constraint. Generated code raises.
CREATE TABLE category (
    id            bigserial PRIMARY KEY,
    scheme        text NOT NULL,
    code          text NOT NULL,
    parent_scheme text,
    parent_code   text,
    CONSTRAINT category_scheme_code_uniq UNIQUE (scheme, code),
    CONSTRAINT category_parent_fkey FOREIGN KEY (parent_scheme, parent_code)
        REFERENCES category (scheme, code)
);
```

```
sqlacodegen --generator declarative postgresql+psycopg2://.../postgres
```

### Generated output

Table A, correct — the foreign key targets `id`, and `remote_side` says `id`:

```python
class Node(Base):
    ...
    parent: Mapped[Optional['Node']] = relationship('Node', remote_side=[id], back_populates='parent_reverse')
    parent_reverse: Mapped[list['Node']] = relationship('Node', remote_side=[parent_id], back_populates='parent')
```

Table B, incorrect — the foreign key targets `(scheme, code)`, but `remote_side` still says `id`:

```python
class Category(Base):
    ...
    category: Mapped[Optional['Category']] = relationship('Category', remote_side=[id], back_populates='category_reverse')
    category_reverse: Mapped[list['Category']] = relationship('Category', remote_side=[parent_scheme, parent_code], back_populates='category')
```

`id` takes no part in that relationship. The many-to-one side should be `remote_side=[scheme, code]`.

### Result

```python
from sqlalchemy.orm import configure_mappers
import models          # imports fine

configure_mappers()
```

```
sqlalchemy.exc.ArgumentError: Relationship Category.category could not determine any
unambiguous local/remote column pairs based on join condition and remote_side arguments.
Consider using the remote() annotation to accurately mark those elements of the join
condition that are on the remote side of the relationship.
```

Table A configures cleanly under the same conditions.

Worth noting for anyone hitting this: the failure is easy to miss in a smoke test, because importing the module succeeds and even `select(Category.id)` succeeds. Selecting a non-primary-key column, or calling `configure_mappers()`, is what surfaces it.

### Expected

`remote_side` on the many-to-one side derived from the foreign key's referenced columns rather than from the primary key. The two coincide only when the self-reference targets the primary key, which is why table A is unaffected.

NixBiks:
## Changes

Fixes #484.

When a table has a self-referential foreign key that targets a `UNIQUE` constraint instead of the primary key, the generated `remote_side` on the many-to-one side named the primary key columns. Those columns take no part in the join, so mapper configuration failed:

```
sqlalchemy.exc.ArgumentError: Relationship Category.category could not determine any
unambiguous local/remote column pairs based on join condition and remote_side arguments.
```

The generated module still imports cleanly, so an import-only smoke test does not catch this; the error only appears on the first query or on `configure_mappers()`.

The cause is in `DeclarativeGenerator.generate_relationships`, which built `remote_side` from `constraint.referred_table.primary_key` rather than from the columns the foreign key refers to. It now uses `constraint.elements`, so `remote_side` names the referred columns. Where the foreign key does target the primary key the two are the same, and the output does not change.

## Checklist

- [x] You've added tests (in `tests/`) which would fail without your patch — `test_onetomany_selfref_composite_unique_target` in `tests/test_generator_declarative.py`, covering a composite self-referential foreign key against a `UNIQUE` constraint. It fails on `master` with `remote_side=[id]` and passes with the fix. Full suite: 158 passed; pre-commit clean.


sheinbergon (approved):

sheinbergon:
10x @NixBiks

agronholm:
Thanks!

